PR #18, opened 2014-04-25
Multiclass CSP

mbillingr:
It is not clear how to optimally select components in the multiclass case.
This will require some further investigation...

### DIFF
--- a/scot/builtin/csp.py
+++ b/scot/builtin/csp.py
@@ -5,9 +5,9 @@
 """common spatial patterns (CSP) implementation"""
 
 import numpy as np
-from scipy.linalg import eigh
+from scipy.linalg import eig
 from itertools import combinations
-    
+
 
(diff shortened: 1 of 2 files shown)


cle1109:
Recently I modified the CSP implementation in mne-python. We're using `eigh` there because if that fails we know something's wrong. What's wrong with using it here as well?

mbillingr:
I don't know about the eig/eigh thing, here. Maybe this was an old compatibility.

Anyway, I've never finished this multiclass CSP thing. I can't remember into which issue I ran exactly, but there were some conceptual open questions that need to be thought through.

cle1109:
OK - how do we proceed with this PR then? I kind of don't want to have stale PRs hanging around that will probably never be finished. I think it's better to collect ideas in issues.

mbillingr:
Let's close this PR. I've still got the branch in my fork.

PR #30, opened 2014-06-23
ENH: parallelization

mbillingr:


mbillingr:
Should we add this to 0.2? I don't think there are any outstanding issues.

cle1109:
Sure - if it works feel free to add the tag and merge.

mbillingr:
Maybe it's easier if we merge the transpose branch first (when it's done) and then adjust this PR to the new data format. Otherwise I fear we'll inflate the transpose branch more and more.

cle1109:
Agreed.

mbillingr:
rebased to master.

TODO: 
 [ ] guard examples with proper `if __name__ == '__main__':` so that they can utilize parallel computations.
 [ ] set `n_jobs` in examples, or default to -1?

cle1109:
Which examples do you mean?

mbillingr:
All of them, basically.
The ones that use whiteness test, in particular.

cle1109:
I don't understand why you need this if statement in the examples - all examples are scripts meant to be run and not imported anyway.

mbillingr:
I think it's good practice in general.

Specifically, joblib needs to import functions from the main module to run them in the worker processes. If you do not guard the script, it runs again every time a process tries to import from it. It then creates workers on its' own that do the same thing again. In short, processes are spawned endlessly.

cle1109:
OK - but then this is relevant information for users, who also have to wrap their script around `if __name__ == '__main__':`, right? If that's the case, I suggest that we make the parallel execution purely optional and clearly state in the docstring how the parallel version must be used.

mbillingr:
Parallel execution is optional already. It's a good idea to mention that in the docstring.

Also, joblib does a decent job in detecting this situation and spits out a warning thet tells the user to wrap their script.

cle1109:
Good. But I'm not sure if we should wrap all examples. This makes them harder to read. IMO we should create one specific example that shows how to perform the whiteness test in parallel.

cle1109:
Also, you could reduce the boilerplate a bit by directly calling the function of interest instead of wrapping this in a `main` function with a single line in its body (as e.g. in `test_var_builtin.py`).

mbillingr:
I don't think one extra line would make the examples harder to read, but it would be nice if they ran faster due to parallelization.

Now that you mention it - this aberration in `test_var_builtin.py` is an old relict of times when I had no idea about unit testing. It will be reomved immediately.

cle1109:
True, one extra line isn't that bad. Still, I don't want all of our examples wrapped, because users might ask themselves why this is necessary. After all, parallelization is an optional feature. Can we agree that we only use parallelization on our slow examples?

mbillingr:
If they ask themselves and start investigating they might find out that it is good practice to wrap the main part of the code, and we did them a favor in the end :)

Let's decide later if/which examples to parallelize. For now we can simply create a parallelization example and leave the others untouched.

cle1109:
Sounds good. I suggest to parallelize `validation.py` - or do you want to create a new example?

mbillingr:
I want to create a new example where automatic model order detection and whiteness test are parallelized. We could also time the difference.

cle1109:
Good idea. We should probably rename this PR because the changes already go beyond just adding a parallelized whiteness test.

mbillingr:
Something strange is happening in the parallelization:

```
n_jobs:   -1, optimization: 0.23s, whiteness: 0.90s
n_jobs: None, optimization: 9.60s, whiteness: 33.81s
n_jobs:    1, optimization: 9.60s, whiteness: 33.75s
n_jobs:    2, optimization: 0.22s, whiteness: 0.92s
n_jobs:    3, optimization: 0.18s, whiteness: 0.92s
n_jobs:    4, optimization: 0.20s, whiteness: 0.90s
n_jobs:    5, optimization: 0.11s, whiteness: 0.92s
n_jobs:    6, optimization: 0.12s, whiteness: 0.92s
n_jobs:    7, optimization: 0.14s, whiteness: 0.93s
n_jobs:    8, optimization: 0.16s, whiteness: 0.92s
```
(n_jobs=-1 means *automatic* and is equivalent to n_jobs=4)

No matter how many jobs are used, the computation is more or less equally fast, unless parallelization is disabled - in this case computation is much smaller.

These results do not make sense. I could expect computation to take roughly half the time when going from 1 to two jobs. Furthermore, I would expect continuous improvement for up to four jobs (number of cores in my system).

@cle1109 could you run the example to verify this behavior, please?

cle1109:
Yes, this is strange. Here's the output on my machine:
```
n_jobs:   -1, optimization: 0.86s, whiteness: 4.32s
n_jobs: None, optimization: 0.83s, whiteness: 4.22s
n_jobs:    1, optimization: 0.84s, whiteness: 4.21s
n_jobs:    2, optimization: 0.83s, whiteness: 4.21s
n_jobs:    3, optimization: 0.84s, whiteness: 4.23s
n_jobs:    4, optimization: 0.83s, whiteness: 4.23s
n_jobs:    5, optimization: 0.84s, whiteness: 4.40s
n_jobs:    6, optimization: 0.89s, whiteness: 4.40s
n_jobs:    7, optimization: 0.87s, whiteness: 4.63s
n_jobs:    8, optimization: 0.84s, whiteness: 4.60s
```

cle1109:
Can you rule out caching/memoization?

mbillingr:
I don't think it's caching because the timing does not depend on the order.

Did you check CPU utilization - do the computations really run in parallel on your machine? It's strange that you see absolutely no difference.

Ah, well! I assume you don't have `joblib` installed. I've written a fallback that tries to use sklearn, but that's not included in this branch, yet.

cle1109:
> Did you check CPU utilization - do the computations really run in parallel on your machine? It's strange that you see absolutely no difference.

Hm. As far as I can tell, there's no difference in CPU load. It seems like the load gets distributed among all available cores. It's really hard to tell the difference using htop or OS X's activity monitor.

> Ah, well! I assume you don't have joblib installed. I've written a fallback that tries to use sklearn, but that's not included in this branch, yet.

Actually, `joblib` 0.9.4 is installed.

> That makes me think... we decided to make parallelization explicit. So it makes no longer sense to silently fall back to serial execution. It may be better to raise an error (or at least a warning) if the user tries to set n_jobs without having joblib available.

Yes, that's a good idea. I would raise an error if `joblib` (or `sklearn`) is not available.

cle1109:
On a related note:

```
n_jobs:   -1, optimization: 0.80s, whiteness: 4.06s, p = 0.85
n_jobs: None, optimization: 0.81s, whiteness: 4.11s, p = 0.74
n_jobs:    1, optimization: 0.84s, whiteness: 4.26s, p = 0.82
n_jobs:    2, optimization: 0.81s, whiteness: 4.08s, p = 0.83
n_jobs:    3, optimization: 0.83s, whiteness: 4.06s, p = 0.83
n_jobs:    4, optimization: 0.80s, whiteness: 4.04s, p = 0.83
n_jobs:    5, optimization: 0.83s, whiteness: 4.04s, p = 0.79
n_jobs:    6, optimization: 0.81s, whiteness: 4.02s, p = 0.79
n_jobs:    7, optimization: 0.82s, whiteness: 4.04s, p = 0.79
n_jobs:    8, optimization: 0.81s, whiteness: 4.12s, p = 0.81
```

I set the model order to 5 and printed the p-value of the whiteness test. Why are the p-values different?

mbillingr:
> I set the model order to 5 and printed the p-value of the whiteness test. Why are the p-values different?

Very interesting question... I'll look into it.

cle1109:
I think we should remove this PR from the 0.2 release - obviously, there are some problems that need to be resolved first. What do you think? I would prefer to release 0.2 ASAP.

mbillingr:
That's fine with me.

cle1109:
Can you rebase please?

mbillingr:
1. The *p*-values are different because we only set the random seed at the beginning of the script, so each run works with different permutations of the data. When I put the seed into the loop all *p*-values are the same.

2. The timing was totally wrong. Apparently, `time.clock()` is deprecated and it reported far too small values. `time.perf_counter()` seems to do a better job:

```
n_jobs:   -1, optimization: 11.61s, whiteness: 23.03s, p = 0.123
n_jobs: None, optimization: 15.54s, whiteness: 37.07s, p = 0.123
n_jobs:    1, optimization: 16.79s, whiteness: 37.27s, p = 0.123
n_jobs:    2, optimization: 11.65s, whiteness: 23.24s, p = 0.123
n_jobs:    3, optimization: 16.04s, whiteness: 22.32s, p = 0.123
n_jobs:    4, optimization: 11.50s, whiteness: 21.83s, p = 0.123
n_jobs:    5, optimization: 20.63s, whiteness: 22.24s, p = 0.123
n_jobs:    6, optimization: 34.17s, whiteness: 22.12s, p = 0.123
n_jobs:    7, optimization: 53.74s, whiteness: 21.88s, p = 0.123
n_jobs:    8, optimization: 11.38s, whiteness: 22.56s, p = 0.123
```
I suppose my laptop only has two cores, that's why there is no speed improvement for more than two jobs.

cle1109:
This looks really good! I'll try it tomorrow on my laptop. Concerning your results, I guess a sane default for most people will be 2 parallel jobs, unless we want to determine the number of cores (I wonder if such an automatic detection is already available in joblib?).

mbillingr:
> I wonder if such an automatic detection is already available in joblib?

`n_jobs=-1`, but it does not tell you the actual number of jobs used.

cle1109:
Python crashes in line 56 of the example - any idea what's going on?

cle1109:
Seems to be related to my Python setup - it does not crash when using Anaconda Python. However, this is definitely not good. In addition, the reported times do not depend on the number of jobs:

```
n_jobs:   -1, optimization: 1.34s, whiteness: 6.81s, p = 0.123
n_jobs: None, optimization: 1.33s, whiteness: 8.59s, p = 0.123
n_jobs:    1, optimization: 1.21s, whiteness: 8.69s, p = 0.123
n_jobs:    2, optimization: 1.45s, whiteness: 7.97s, p = 0.123
n_jobs:    3, optimization: 2.02s, whiteness: 8.79s, p = 0.123
n_jobs:    4, optimization: 1.39s, whiteness: 8.22s, p = 0.123
n_jobs:    5, optimization: 1.81s, whiteness: 7.73s, p = 0.123
n_jobs:    6, optimization: 2.42s, whiteness: 7.67s, p = 0.123
n_jobs:    7, optimization: 3.68s, whiteness: 7.90s, p = 0.123
n_jobs:    8, optimization: 1.16s, whiteness: 7.38s, p = 0.123
```

cle1109:
And here the output from my Windows desktop machine:

```
n_jobs:   -1, optimization: 4.24s, whiteness: 6.53s, p = 0.123
n_jobs: None, optimization: 1.39s, whiteness: 8.76s, p = 0.123
n_jobs:    1, optimization: 1.25s, whiteness: 8.80s, p = 0.123
n_jobs:    2, optimization: 4.99s, whiteness: 7.36s, p = 0.123
n_jobs:    3, optimization: 4.26s, whiteness: 6.46s, p = 0.123
n_jobs:    4, optimization: 2.99s, whiteness: 6.23s, p = 0.123
n_jobs:    5, optimization: 4.04s, whiteness: 6.65s, p = 0.123
n_jobs:    6, optimization: 5.35s, whiteness: 7.05s, p = 0.123
n_jobs:    7, optimization: 6.96s, whiteness: 7.38s, p = 0.123
n_jobs:    8, optimization: 3.06s, whiteness: 7.51s, p = 0.123
```

Doesn't make sense.

mbillingr:
I have added some debug output. Could you please enable verbosity in your configuration and run the example again?

cle1109:
```
/usr/local/lib/python3.5/site-packages/scipy/linalg/basic.py:884: RuntimeWarning: internal gelsd driver lwork query error, required iwork dimension not returned. This is likely the result of LAPACK bug 0038, fixed in LAPACK 3.2.2 (released July 21, 2010). Falling back to 'gelss' driver.
  warnings.warn(mesg, RuntimeWarning)
False
attempting to import joblib
joblib imported successfully
```

Then it crashes.

mbillingr:
oh.. no idea..

Any idea how to debug parallel code?

cle1109:
No. I tracked it down to line 579 in joblib's `parallel.py`:

    job = self._pool.apply_async(SafeFunction(batch), callback=cb)

I don't know how to debug this further.

mbillingr:
This sounds like a problem with joblib.

cle1109:
Yes. But in addition, it doesn't seem to work on my machine even if it doesn't crash.

mbillingr:
What's the output if it doesn't crash?

cle1109:
See my comments above (laptop and local Windows desktop).

mbillingr:
Can you run them again with debugging output?

cle1109:
```
Registering scot backend: builtin
Registering scot backend: sklearn
Activating scot backend: builtin
True
attempting to import joblib
joblib imported successfully
running  <function _get_msge_with_gradient at 0x10ebd80d0>  in parallel
[Parallel(n_jobs=-1)]: Done   1 out of   4 | elapsed:    0.3s remaining:    0.9s
[Parallel(n_jobs=-1)]: Done   4 out of   4 | elapsed:    0.4s finished
[Parallel(n_jobs=-1)]: Done   1 out of   4 | elapsed:    0.5s remaining:    1.4s
[Parallel(n_jobs=-1)]: Done   4 out of   4 | elapsed:    0.8s finished
attempting to import joblib
joblib imported successfully
running  <function _calc_q_statistic at 0x10ebd1b70>  in parallel
[Parallel(n_jobs=-1)]: Done   1 out of 130 | elapsed:    0.4s remaining:   47.9s
[Parallel(n_jobs=-1)]: Done 1000 out of 1000 | elapsed:    7.3s finished
n_jobs:   -1, optimization: 1.33s, whiteness: 7.39s, p = 0.123
running  <function _get_msge_with_gradient at 0x10ebd80d0>  serially
running  <function _calc_q_statistic at 0x10ebd1b70>  serially
n_jobs: None, optimization: 1.24s, whiteness: 8.59s, p = 0.123
attempting to import joblib
joblib imported successfully
running  <function _get_msge_with_gradient at 0x10ebd80d0>  in parallel
[Parallel(n_jobs=1)]: Done   1 jobs       | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done   1 out of   1 | elapsed:    0.0s finished
[Parallel(n_jobs=1)]: Done   1 jobs       | elapsed:    0.1s
[Parallel(n_jobs=1)]: Done   1 out of   1 | elapsed:    0.1s finished
[Parallel(n_jobs=1)]: Done   1 jobs       | elapsed:    0.1s
[Parallel(n_jobs=1)]: Done   1 out of   1 | elapsed:    0.1s finished
[Parallel(n_jobs=1)]: Done   1 jobs       | elapsed:    0.1s
[Parallel(n_jobs=1)]: Done   1 out of   1 | elapsed:    0.1s finished
[Parallel(n_jobs=1)]: Done   1 jobs       | elapsed:    0.2s
[Parallel(n_jobs=1)]: Done   1 out of   1 | elapsed:    0.2s finished
[Parallel(n_jobs=1)]: Done   1 jobs       | elapsed:    0.2s
[Parallel(n_jobs=1)]: Done   1 out of   1 | elapsed:    0.2s finished
[Parallel(n_jobs=1)]: Done   1 jobs       | elapsed:    0.2s
[Parallel(n_jobs=1)]: Done   1 out of   1 | elapsed:    0.2s finished
[Parallel(n_jobs=1)]: Done   1 jobs       | elapsed:    0.3s
[Parallel(n_jobs=1)]: Done   1 out of   1 | elapsed:    0.3s finished
attempting to import joblib
joblib imported successfully
running  <function _calc_q_statistic at 0x10ebd1b70>  in parallel
[Parallel(n_jobs=1)]: Done   1 jobs       | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done  50 jobs       | elapsed:    0.5s
[Parallel(n_jobs=1)]: Done 200 jobs       | elapsed:    2.2s
[Parallel(n_jobs=1)]: Done 450 jobs       | elapsed:    4.8s
[Parallel(n_jobs=1)]: Done 800 jobs       | elapsed:    8.2s
[Parallel(n_jobs=1)]: Done 1000 out of 1000 | elapsed:   10.1s finished
n_jobs:    1, optimization: 1.27s, whiteness: 10.12s, p = 0.123
attempting to import joblib
joblib imported successfully
running  <function _get_msge_with_gradient at 0x10ebd80d0>  in parallel
[Parallel(n_jobs=2)]: Done   1 out of   2 | elapsed:    0.1s remaining:    0.1s
[Parallel(n_jobs=2)]: Done   2 out of   2 | elapsed:    0.1s finished
[Parallel(n_jobs=2)]: Done   1 out of   2 | elapsed:    0.2s remaining:    0.2s
[Parallel(n_jobs=2)]: Done   2 out of   2 | elapsed:    0.3s finished
[Parallel(n_jobs=2)]: Done   1 out of   2 | elapsed:    0.3s remaining:    0.3s
[Parallel(n_jobs=2)]: Done   2 out of   2 | elapsed:    0.3s finished
[Parallel(n_jobs=2)]: Done   1 out of   2 | elapsed:    0.4s remaining:    0.4s
[Parallel(n_jobs=2)]: Done   2 out of   2 | elapsed:    0.5s finished
attempting to import joblib
joblib imported successfully
running  <function _calc_q_statistic at 0x10ebd1b70>  in parallel
[Parallel(n_jobs=2)]: Done   1 out of  11 | elapsed:    0.0s remaining:    0.4s
[Parallel(n_jobs=2)]: Done 1000 out of 1000 | elapsed:    7.7s finished
n_jobs:    2, optimization: 1.67s, whiteness: 7.76s, p = 0.123
attempting to import joblib
joblib imported successfully
running  <function _get_msge_with_gradient at 0x10ebd80d0>  in parallel
[Parallel(n_jobs=3)]: Done   1 out of   3 | elapsed:    0.3s remaining:    0.5s
[Parallel(n_jobs=3)]: Done   3 out of   3 | elapsed:    0.3s finished
[Parallel(n_jobs=3)]: Done   1 out of   3 | elapsed:    0.4s remaining:    0.8s
[Parallel(n_jobs=3)]: Done   3 out of   3 | elapsed:    0.6s finished
[Parallel(n_jobs=3)]: Done   1 out of   3 | elapsed:    0.9s remaining:    1.8s
[Parallel(n_jobs=3)]: Done   3 out of   3 | elapsed:    1.1s finished
attempting to import joblib
joblib imported successfully
running  <function _calc_q_statistic at 0x10ebd1b70>  in parallel
[Parallel(n_jobs=3)]: Done   1 out of   6 | elapsed:    0.1s remaining:    0.3s
[Parallel(n_jobs=3)]: Done 1000 out of 1000 | elapsed:    9.2s finished
n_jobs:    3, optimization: 2.27s, whiteness: 9.36s, p = 0.123
attempting to import joblib
joblib imported successfully
running  <function _get_msge_with_gradient at 0x10ebd80d0>  in parallel
[Parallel(n_jobs=4)]: Done   1 out of   4 | elapsed:    0.3s remaining:    0.9s
[Parallel(n_jobs=4)]: Done   4 out of   4 | elapsed:    0.4s finished
[Parallel(n_jobs=4)]: Done   1 out of   4 | elapsed:    0.5s remaining:    1.4s
[Parallel(n_jobs=4)]: Done   4 out of   4 | elapsed:    0.8s finished
attempting to import joblib
joblib imported successfully
running  <function _calc_q_statistic at 0x10ebd1b70>  in parallel
[Parallel(n_jobs=4)]: Done   1 out of  78 | elapsed:    0.2s remaining:   18.6s
[Parallel(n_jobs=4)]: Done 1000 out of 1000 | elapsed:    8.1s finished
n_jobs:    4, optimization: 1.51s, whiteness: 8.17s, p = 0.123
attempting to import joblib
joblib imported successfully
running  <function _get_msge_with_gradient at 0x10ebd80d0>  in parallel
[Parallel(n_jobs=5)]: Done   1 out of   5 | elapsed:    0.3s remaining:    1.2s
[Parallel(n_jobs=5)]: Done   5 out of   5 | elapsed:    0.5s finished
[Parallel(n_jobs=5)]: Done   1 out of   5 | elapsed:    0.8s remaining:    3.1s
[Parallel(n_jobs=5)]: Done   5 out of   5 | elapsed:    1.3s finished
attempting to import joblib
joblib imported successfully
running  <function _calc_q_statistic at 0x10ebd1b70>  in parallel
[Parallel(n_jobs=5)]: Done   1 out of  54 | elapsed:    0.2s remaining:   12.9s
[Parallel(n_jobs=5)]: Done 1000 out of 1000 | elapsed:    7.7s finished
n_jobs:    5, optimization: 2.04s, whiteness: 7.75s, p = 0.123
attempting to import joblib
joblib imported successfully
running  <function _get_msge_with_gradient at 0x10ebd80d0>  in parallel
[Parallel(n_jobs=6)]: Done   1 out of   6 | elapsed:    0.3s remaining:    1.6s
[Parallel(n_jobs=6)]: Done   6 out of   6 | elapsed:    0.6s finished
[Parallel(n_jobs=6)]: Done   1 out of   6 | elapsed:    1.4s remaining:    6.8s
[Parallel(n_jobs=6)]: Done   6 out of   6 | elapsed:    1.9s finished
attempting to import joblib
joblib imported successfully
running  <function _calc_q_statistic at 0x10ebd1b70>  in parallel
[Parallel(n_jobs=6)]: Done   1 out of  94 | elapsed:    0.4s remaining:   33.1s
[Parallel(n_jobs=6)]: Done 1000 out of 1000 | elapsed:    7.6s finished
n_jobs:    6, optimization: 2.67s, whiteness: 7.68s, p = 0.123
attempting to import joblib
joblib imported successfully
running  <function _get_msge_with_gradient at 0x10ebd80d0>  in parallel
[Parallel(n_jobs=7)]: Done   1 out of   7 | elapsed:    0.4s remaining:    2.3s
[Parallel(n_jobs=7)]: Done   7 out of   7 | elapsed:    0.8s finished
[Parallel(n_jobs=7)]: Done   1 out of   7 | elapsed:    1.9s remaining:   11.5s
[Parallel(n_jobs=7)]: Done   7 out of   7 | elapsed:    2.8s finished
attempting to import joblib
joblib imported successfully
running  <function _calc_q_statistic at 0x10ebd1b70>  in parallel
[Parallel(n_jobs=7)]: Done   1 out of  56 | elapsed:    0.3s remaining:   16.2s
[Parallel(n_jobs=7)]: Done 1000 out of 1000 | elapsed:    7.5s finished
n_jobs:    7, optimization: 3.89s, whiteness: 7.53s, p = 0.123
attempting to import joblib
joblib imported successfully
running  <function _get_msge_with_gradient at 0x10ebd80d0>  in parallel
[Parallel(n_jobs=8)]: Done   1 out of   8 | elapsed:    0.4s remaining:    2.5s
[Parallel(n_jobs=8)]: Done   8 out of   8 | elapsed:    1.1s finished
attempting to import joblib
joblib imported successfully
running  <function _calc_q_statistic at 0x10ebd1b70>  in parallel
[Parallel(n_jobs=8)]: Done   1 out of  92 | elapsed:    0.3s remaining:   31.8s
[Parallel(n_jobs=8)]: Done 1000 out of 1000 | elapsed:    8.4s finished
n_jobs:    8, optimization: 1.18s, whiteness: 8.59s, p = 0.123
```

mbillingr:
Hmm.. jbolib seems to be working correctly but there is not much improvement in speed: whiteness runs in 10s with one job and in about 8+/-1s with more than one job.

I assume the system was not busy with background threads that interfered with the timing?

Are you sure the computer has 4 cores? (My laptop also reports 4 cores, but actually there are only 2 cores with hyper-threading.)

cle1109:
> Hmm.. jbolib seems to be working correctly but there is not much improvement in speed: whiteness runs in 10s with one job and in about 8+/-1s with more than one job.

You think the output is normal? Why does it only show completion of the first and the last job (i.e. for 8 jobs 1 out of 8 and 8 out of 8 but nothing in between)?

> I assume the system was not busy with background threads that interfered with the timing?

Nope.

> Are you sure the computer has 4 cores? (My laptop also reports 4 cores, but actually there are only 2 cores with hyper-threading.)

http://ark.intel.com/products/83505/Intel-Core-i7-4770HQ-Processor-6M-Cache-up-to-3_40-GHz

cle1109:
The crash was solved in https://github.com/joblib/joblib/issues/314 - on OS X, the default method of forking doesn't work. Therefore, the environment variable `JOBLIB_START_METHOD` needs to be set to `forkserver` (instead of the default `fork`).

Here's the output:

```
n_jobs:   -1, optimization: 4.42s, whiteness: 4.95s, p = 0.123
n_jobs: None, optimization: 1.43s, whiteness: 10.78s, p = 0.123
n_jobs:    1, optimization: 1.44s, whiteness: 10.76s, p = 0.123
n_jobs:    2, optimization: 5.22s, whiteness: 7.60s, p = 0.123
n_jobs:    3, optimization: 4.43s, whiteness: 6.47s, p = 0.123
n_jobs:    4, optimization: 3.52s, whiteness: 5.55s, p = 0.123
n_jobs:    5, optimization: 4.12s, whiteness: 6.31s, p = 0.123
n_jobs:    6, optimization: 6.19s, whiteness: 5.26s, p = 0.123
n_jobs:    7, optimization: 6.25s, whiteness: 5.14s, p = 0.123
n_jobs:    8, optimization: 2.65s, whiteness: 5.37s, p = 0.123
```

The whiteness test decreases until 4 jobs, this makes sense. However, the optimization is fastest for 1 job.

I don't know, this seems very arbitrary and inconsistent. As much as I'd like to speed up computations, I don't think joblib is the right choice here. We should probably look into other options, such as Cython (no idea if independent loop iterations can be sped up with Cython) or Numba (which is notoriously difficult to set up under non-Anaconda envs).

mbillingr:
Nevermind optimization for the moment - I have implemented this in a very weird way as far as joblib is concerned (it uses a while loop); we should address this in another PR.

The whiteness test is ideal for joblib because of the independent loop iterations and the speedup seems ok (a factor 2 is a bit low with four cores on your system, but better than nothing)

Numpy could be an option, but it makes development and deployment more difficult.

cle1109:
> Nevermind optimization for the moment - I have implemented this in a very weird way as far as joblib is concerned (it uses a while loop); we should address this in another PR.

OK.

> The whiteness test is ideal for joblib because of the independent loop iterations and the speedup seems ok (a factor 2 is a bit low with four cores on your system, but better than nothing)

Maybe this relatively unimpressive speedup is because I have to use this other forking method in joblib. I agree it's better than nothing, but still the whole joblib package doesn't convince me in our setup.

> Numpy could be an option, but it makes development and deployment more difficult.

You mean Numba or Numpy? Can we optimize our Numpy code (e.g. vectorization)?

What about Cython? Scikit-learn is using Cython in performance-critical parts.

mbillingr:
> but still the whole joblib package doesn't convince me in our setup.

I don't quite understand. You seem to insist on running code serially. But whatever speedup we achieve by other methods (see below), we can still get an additional factor of at least 2 when we add parallelization.

> What about Cython? Scikit-learn is using Cython in performance-critical parts.

I meant Cython (I've edited the post, but it looks like you have seen the old version).

If we use Cython we can no longer do simple releases. Instead, we have to build binary packages for every supported platform. Also, as a developer you have to install the package (or at least build it) on your local system before you can work on it. If you pull in changes in the Cython part of the code you have to remember to rebuild. Finally, Cython code is slightly more difficult to maintain as you have to care about typed variables.

Numba is even worse. It shifts the burden of maintining a compilation environment from the developer to the user.

Don't get me wrong. Both, cython and numba, are great tools that I'd like to use in code for my personal use. We may even be able to get nice speedups in SCoT, too. However, I think they would introduce too much maintenance cost to our pet project.

cle1109:
> I don't quite understand. You seem to insist on running code serially. But whatever speedup we achieve by other methods (see below), we can still get an additional factor of at least 2 when we add parallelization.

If we include support for parallel jobs in SCoT, I expect this to (1) run out of the box and (2) produce significant speedup. Otherwise, it's not worth the additional work.

Concerning (1), joblib crashes on Mac OS X by default. We would need to take appropriate measures (i.e. set the environment variable or maybe this setting can be invoked inside the joblib module) in order to fix this. I don't think that it is up to SCoT to fix this though - it would add some complexity to our codebase (checking for OS, change the setting), and I really don't want to do this unless the benefit outweighs the costs. Ideally, I think joblib should choose the correct settings depending on which platform it is run, so we could ask the devs what they think about that.

Re (2), although a speedup of 2 is certainly great, it still leaves me feeling that there's something not quite right. Why do you see an almost linear speed increase on your machine, whereas I don't? I just tested the example on my Windows quad-core running the latest Anaconda, and the behavior is even worse there (factor 1.44):

```
n_jobs:   -1, optimization: 3.18s, whiteness: 5.91s, p = 0.123
n_jobs: None, optimization: 1.61s, whiteness: 8.51s, p = 0.123
n_jobs:    1, optimization: 1.25s, whiteness: 8.58s, p = 0.123
n_jobs:    2, optimization: 4.99s, whiteness: 6.79s, p = 0.123
n_jobs:    3, optimization: 4.26s, whiteness: 5.82s, p = 0.123
n_jobs:    4, optimization: 3.07s, whiteness: 5.88s, p = 0.123
n_jobs:    5, optimization: 4.05s, whiteness: 6.16s, p = 0.123
n_jobs:    6, optimization: 7.24s, whiteness: 7.05s, p = 0.123
n_jobs:    7, optimization: 7.92s, whiteness: 7.23s, p = 0.123
n_jobs:    8, optimization: 3.08s, whiteness: 8.81s, p = 0.123
```

> I meant Cython (I've edited the post, but it looks like you have seen the old version).

You're right about Cython. Let's stick to pure Python.

It would be nice though to make it easy for the user to optimize/parallelize certain functions such as the whiteness test without having to modify our code. Is it possible to apply such optimizations in code that only calls SCoT routines? Probably not unless we create a function/generator that returns/yields one iteration at a time, which the user then needs to call in a loop... :-/

mbillingr:
> If we include support for parallel jobs in SCoT, I expect this to (1) run out of the box and (2) produce significant speedup. Otherwise, it's not worth the additional work.

I agree. What is a significant speedup? :)
Isn't an optional performance feature that can give improved results for some users/systems good enough?

> I don't think that it is up to SCoT to fix this though

I totally agree. First, if we fixed it and then they fixed it, we'd have to fix our fix again...
Second, if the feature is optional and the user wants to activate it they can also deal with the peculiarities of their systems (IMO).

>  Why do you see an almost linear speed increase on your machine, whereas I don't? I just tested the example on my Windows quad-core running the latest Anaconda, and the behavior is even worse there (factor 1.44)

I get even more puzzling behavior on Windows/Anaconda, here: same performance for `n_jobs <=4` and much worse performance for `n_jobs > 4`...
The explanation simple. Have you checked the Task Manager? When I run with only one job there are four busy cores. This means there is another parallelization going on in the background. Most likely Anaconda is using a parallelized version of BLAS/LAPACK (MKL?). If one job already makes optimal use of the processor cores, performance won't improve by adding more jobs. My Linux machine with good old BLAS, in constrast, gets most out of Joblib.

> It would be nice though to make it easy for the user to optimize/parallelize certain functions such as the whiteness test without having to modify our code. Is it possible to apply such optimizations in code that only calls SCoT routines?

I don't think it's possible because the looping is done internally in our routines.
Hmm.. modifiying our code to expose loops would be very invasive to the API (and it won't exactly happen without having to modify our code).

What's wrong with the current implementation where users can optionally enable parallelization? Nobody is forced to use it if it does not give a benefit on one's system.

cle1109:
> What's wrong with the current implementation where users can optionally enable parallelization? Nobody is forced to use it if it does not give a benefit on one's system.

Of course. The thing is that it will likely only work for a minority of users. This sucks - I want to use it too :-)!

Most people will be using Anaconda Python, which means they will use multithreaded BLAS/LAPACK libs. It could even be the case that it is faster to run only one job, but rely on multithreaded linalg computations.

So all I'm saying is that we already have a lot of code that does very little for most users. However, if a user like me sees that there are these options, it is likely that they try it and be disappointed.

Finally, I guess our code is very specifically tailored towards joblib. Or can joblib be replaced by something else?

mbillingr:
> So all I'm saying is that we already have a lot of code that does very little for most users. However, if a user like me sees that there are these options, it is likely that they try it and be disappointed.

I'm not sure if this is the case for most users and if it really is that much code.
Anyway, I tend to believe it is better to be disappointed because resources are already used optimally than to have someone (me! :p) stuck with slow computations although there is optimization potential.

> Finally, I guess our code is very specifically tailored towards joblib. Or can joblib be replaced by something else?

Our `parallel.py` is 38 lines long. It implements the function `parallel_loop` that is used for parallelizing list comprehensions. Although this basically is a wrapper around `joblib.Parallel` It should be possible to plug in something else without changing any other part of the code.

cle1109:
> I'm not sure if this is the case for most users and if it really is that much code

Of course I don't know either, but I strongly suspect that the average scientific Python user will be using Anaconda. Which is a scenario where our parallelization doesn't really work well.

I agree that it's not that much code that's affected.

> Anyway, I tend to believe it is better to be disappointed because resources are already used optimally than to have someone (me! :p) stuck with slow computations although there is optimization potential.

Well, yeah, as long as such optimizations are not baked in too much into our toolbox. Plus, you seem to be using Python installed via a package manger from a Linux distribution (Arch?) with a non-multithreaded BLAS/LAPACK - why do I still have the feeling that you're basically the only one benefitting from our parallelization :-)?

> Our `parallel.py` is 38 lines long. It implements the function `parallel_loop` that is used for parallelizing list comprehensions. Although this basically is a wrapper around `joblib.Parallel` It should be possible to plug in something else without changing any other part of the code.

Cool. So I think we're good regarding this PR in general. We need to keep the current limitations in the back of our heads, because I'm sure we will get reports that parallelization isn't working for some people. How do you think we should handle this? At least we should add some notes in the docstrings (of the module and/or functions).

mbillingr:
> Python installed via a package manger from a Linux distribution (Arch?) with a non-multithreaded BLAS/LAPACK - why do I still have the feeling that you're basically the only one benefitting from our parallelization :-)?

Exactly. If I am the only one benefitting from the parallelization that's fine with me. That covers 50% of our userbase :)

> How do you think we should handle this? At least we should add some notes in the docstrings

I'm not sure if docstrings are the best place. After all, the functions just do what they are supposed to do, which is running multiple jobs. The fact that this may not be beneficial in all cases should be documented, but I do not know where.

cle1109:
> Exactly. If I am the only one benefitting from the parallelization that's fine with me. That covers 50% of our userbase :)

:-) - let's just not get to a point where essential functions run only on a specific setup of one user.

> I'm not sure if docstrings are the best place. After all, the functions just do what they are supposed to do, which is running multiple jobs. The fact that this may not be beneficial in all cases should be documented, but I do not know where.

Adding a note in `README.md` would be obvious. However, this somehow over-emphasized our parallelization mechanism, and likely encourages many people to try it out. If we write, on the other hand, that it doesn't work in many cases, this might create a somewhat unprofessional picture of SCoT :-). Don't you think we could add a docstring at the module level in `parallel.py` with this information?

mbillingr:
> Don't you think we could add a docstring at the module level in parallel.py with this information?

We could do that, but this function is never exposed to the user.

cle1109:
Yes, this would require that the user pokes around in the source code. I don't know - this would be a good start and it's better to have it somewhere. Also, shouldn't module-level docstrings be visible in our Sphinx-generated API doc?

mbillingr:
I'd think so... but I don't expect a user to look up the documentation of the parallel module if they want to know about the `n_jobs` argument of the VAR model.

mbillingr:
We could also address such issues in a FAQ.

cle1109:
+1 for FAQ - should we create a FAQ section in `README.md`?

mbillingr:
+1 for FAQ@`README.md`. At least for now. If the FAQ grows in size we can move it to the docs, later.

cle1109:
Yay - tbc at #142.

What about this PR? Is it finished? Maybe you could do less weird things in the optimization example :-). Anything else?

mbillingr:
> Maybe you could do less weird things in the optimization example :-)

What weird things in what example? :)

Some of the debug output should be removed again. Otherwise, I think this PR is done.

cle1109:
You mentioned a few comments above:

> Nevermind optimization for the moment - I have implemented this in a very weird way as far as joblib is concerned (it uses a while loop); we should address this in another PR.

If you want to include this example here, I think this weirdness should be removed first. And of course the debug output should be disabled.

mbillingr:
This weirdness is not due to the example. It is caused by the way parallelism is implemented in order optimization (this is not new).

We have several options:
1. remove order optimization from that example
2. remove order optimization from that example andremove parallelization from order optimization
3. improve parallel order optimization

I'd prefer 3) but it's most work and will take a while. We could also do 1) now and 3) later, in a different PR.

cle1109:
OK, then let's go for 1) now and do 3) later in a separate PR. Can you disable the debug output, rebase, and then merge (if Travis builds pass)?

mbillingr:
Doh. I can't merge because Coveralls is overloaded at the moment and this makes travis believe that the script failed.

cle1109:
Would you mind squashing a few nonessential commits in the meantime?

Issue #42, opened 2015-01-09
Remove binica backend


cle1109:
I think we should remove the `binica` backend because (1) it is not Pythonic, (2) we have an excellent replacement from MNE (which BTW runs much faster anyway), (3) only exists on Linux, and (4) therefore breaks our tests on non-Linux platforms (I just ran the tests on my Mac and 5 tests failed because I don't have the `binica` executable).

If we remove it, this would automatically solve #32 and #37.

kazemakase:
#43

kazemakase:
#43

PR #43, opened 2015-01-09
Remove binica backend

cle1109:
Should be working just fine without binica (addresses #42).


cle1109:
@kazemakase, let me know if you think this PR can be merged.

Issue #41, opened 2015-01-08
Use new-style classes


cle1109:
Would it break anything if scot used new-style classes?

kazemakase:
No, since I believe classes are new-style by default in Python 3.

cle1109:
So we could explicitly derive each class from `object`?

kazemakase:
This is redundant in Python 3, but might improve compatibility with Python >=2.2.

So, yes, I think we should do that eventually.

We don't care that this does not work on Python 2.1 and below, do we?

cle1109:
Right, I think we should focus on current Python versions (3.x and 2.7, probably also 2.6). Definitely not < 2.2. I'll take care of this change in a PR.

kazemakase:
Thank you.

cle1109:
Solved by #45.

PR #45, opened 2015-01-12
Use new-style classes

cle1109:
Use new-style classes, addresses #41.

### DIFF
--- a/scot/connectivity.py
+++ b/scot/connectivity.py
@@ -50,9 +50,9 @@ def connectivity(measure_names, b, c=None, nfft=512):
 
 
 #noinspection PyPep8Naming
-class Connectivity:
+class Connectivity(object):
     """ Calculation of connectivity measures
-    
+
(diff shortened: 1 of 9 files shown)


kazemakase:
Where do all the whitespace changes come from?

cle1109:
Hm, I didn't notice those. Probably PyCharm's PEP8 correction or something. I think spaces on empty lines are [discouraged](http://stackoverflow.com/questions/2727988/python-indentation-in-empty-lines).

kazemakase:
There doesn't seem to be any consensus, but I don't really care about blank lines. They  just make it hard to see the actual changes in the diff.

cle1109:
I agree. Since PyCharm does it this way by default, I'd say we simply adopt this style.

kazemakase:
Strange that It did not do it by default when I wrote that code...

kazemakase:
By the way, I think you can merge this PR anytime.

Issue #46, opened 2015-01-19
Upload to PyPI


cle1109:
Any plans on uploading scot to PyPI?

agramfort:
would be cool

let me know if you need help

cle1109:
I created PR #47 to address this issue - let's continue our discussion over there.

kazemakase:
I think that is an excellent idea, now that we got rid of binica :)

cle1109:
Merged by #47.

PR #47, opened 2015-01-20
Distribute on PyPI

cle1109:
This PR addresses #46. I updated `setup.py`, but I have some questions. For starters, which email address should we use? Ideally, I would like to have an email address where (selected) members of the SCoT development team have access. I don't think GitHub support this?

Any other comments on this file (I used [this](https://github.com/pypa/sampleproject/blob/master/setup.py) as a template).


kazemakase:
I don't think GitHub provides email addresses.

Maybe we can create a gmail (or something like that) address and put the login data somewhere on github where only selected members can see it?

cle1109:
OK. What about creating a mailing list at Google Groups and forward mail to this address to project members? Something like scot-dev@googlegroups.com?

kazemakase:
I like the mailing list idea better than a shared account, but I guess it still requires account management.

cle1109:
At least that's how they do it in the [sample project](https://github.com/pypa/sampleproject/blob/master/setup.py).

cle1109:
OK, I created a collaborative inbox (aka mailling list) on Google Groups. Our email address is scotdev@googlegroups.com. OK?

cle1109:
I uploaded it to [TestPyPI](https://testpypi.python.org/pypi) - can you please check it (as long as it's there)? @agramfort, maybe you can also comment if the package (distribution) is OK like that.

kazemakase:
We should change the description to something along the line of "EEG source connectivity toolbox", because just source connectivity is misleading in a programming context.

setup.py
```diff
@@ -1,17 +1,38 @@
...
+        'Programming Language :: Python :: 2.6',
+        'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.2',
+        'Programming Language :: Python :: 3.3',
+        'Programming Language :: Python :: 3.4',
+    ],
+    keywords='source connectivity EEG MEG ICA',
+    packages=['scot', 'scot.eegtopo', 'scot.external'],
+    install_requires=['numpy >=1.7', 'scipy >=0.12'],
```
agramfort: don't add these deps. It may lead to pip installing numpy and scipy and you don't want this. It's likely to break the install of people.

cle1109: Why don't we want pip installing numpy and scipy? I thought that's exactly what we want, because otherwise scot does not work.

agramfort: import numpy will fail an people are free to install it correctly : apt-get
or conda

pip wheels exist but It's not yet mature to me.

cle1109: OK. On my Mac, I always used pip to install numpy and scipy and never had any problems. I don't know about Linux though, but you are right, on Linux I would also prefer installing these packages through the distribution's package manager and not via pip.

kazemakase: +1 for removing the requirement.

You never know against which blas/lapack libraries pip trys to link numpy (maybe even some included ones). I had big problems with that when trying to distribute my computations over a number of different debian/ubuntu systems in the lab.

cle1109:
I think it would be nice to finally upload scot to PyPI. I removed the dependencies and uploaded the updated version to [TestPyPI](https://testpypi.python.org/pypi/scot/0.1.1). It looks a bit messy, so I guess we need to clean up some files so that the description renders nicely. Also, I had to increase the version number, because TestPyPI wouldn't let me upload an updated version with the same version number.

Any comments?

kazemakase:
I don't litke the style of PyPi anyway, so I guess in that light it looks okay :)

The short description (near the top) appears twice:
> EEG/MEG Source Connectivity Toolbox
EEG/MEG Source Connectivity Toolbox for Python.

Also, according to GRMP, the toolbox is not *for* Python. Instead it is *for* doing connectivity, and written *in* Python...

kazemakase:
I don't like the PyPi style anyway. Guess it's fine the way it is now...

Two comments:
1. Why does the short discription appear twice (near the top)?
> EEG/MEG Source Connectivity Toolbox
EEG/MEG Source Connectivity Toolbox for Python.

2. According to GRMP, the toolbox is not *for* Python. Instead, it is *for* doing connectivity while being written *in* Python... I'm not sure if I agree, though :)

cle1109:
OK - maybe we can copy elements from [scikit-learn](https://pypi.python.org/pypi/scikit-learn/0.16.1)? Their PyPI site looks nice.

Regarding for/in, I don't really care. Let's just drop it and say that SCoT is a Python module for source connectivity estimation. On that note, is it a module or a package :-)?

kazemakase:
[It's a package](http://stackoverflow.com/a/7948504)

kazemakase:
Btw, why does this PR have merge conflicts?

setup.py
```diff
@@ -1,17 +1,37 @@
...
+    author='SCoT Development Team',
+    author_email='scotdev@googlegroups.com',
+    license='MIT',
+    classifiers=[
+        'Development Status :: 4 - Beta',
+        'Intended Audience :: Science/Research',
+        'Topic :: Scientific/Engineering',
+        'License :: OSI Approved :: MIT License',
+        'Programming Language :: Python :: 2',
+        'Programming Language :: Python :: 2.6',
```
kazemakase: I don't think scot works with Python 2.6.

setup.py
```diff
@@ -1,17 +1,37 @@
...
+        'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.2',
+        'Programming Language :: Python :: 3.3',
+        'Programming Language :: Python :: 3.4',
```
kazemakase: Although I'm sure scot works with Python 3.4, we only test on 3.3.
Should we adjust this here?
Should we update our tests?

cle1109: We should support the latest versions of 2 and 3, which are 2.7 and 3.4 at the moment. So yes, let's update our tests.

setup.py
cle1109: OK, we can drop it as long as we keep support for 2.7.

cle1109:
There are conflicts in README.md.

cle1109:
What do you think? https://testpypi.python.org/pypi/scot/0.1.3

kazemakase:
LGTM

cle1109:
Yes, I've already summarized the steps in a local doc. Since this is only relevant for the core maintainers (us two :-)), where should we put it?

kazemakase:
Let's put it under distributions/pypi/.
Is it possible to create a script for automating the process?

cle1109:
Yes, definitely. I will add this script right now.

cle1109:
Is the README sufficient, or do we need a script? It's more or less just 2 lines. Can you try if it works for you (maybe upload it to TestPyPI under a different name)?

Issue #35, opened 2014-10-02
Rename project to "scot"


cle1109:
Should we rename "SCoT" to "scot"? The only reason is that it is much more common to use all lowercase names, and everything else is a bit ugly :-). However, I don't know if this is even possible.

kazemakase:
Actually it broke coveralls...

cle1109:
Very good, thanks for noticing...

kazemakase:
I think it's working again, but let's wait for the next pull request to be sure.

kazemakase:
Hmm.. coveralls seems to have correctly received the results for PR #55: https://coveralls.io/r/scot-dev/scot?branch=master, but the bot did not comment on the PR...

cle1109:
We've had some PRs since then and everything seems to be working - can we close this issue?

kazemakase:
I think so :)

PR #55, opened 2015-03-04
Infomax: Made fixed random seed optional

kazemakase:
This patch makes the forced random seed in Infomax ICA optional.

@cle1109, @jdammers
Does this look good to you?

@dengemann
Maybe this patch is interesting for MNE as well?

### DIFF
--- a/scot/external/infomax_.py
+++ b/scot/external/infomax_.py
@@ -39,7 +39,7 @@ class logger(object):
 def infomax(data, weights=None, l_rate=None, block=None, w_change=1e-12,
             anneal_deg=60., anneal_step=0.9, extended=False, n_subgauss=1,
             kurt_size=6000, ext_blocks=1, max_iter=200,
-            random_state=None, verbose=None):
+            random_state=None, force_random_seed=False, verbose=None):
     """Run the (extended) Infomax ICA decomposition on raw data
 
     based on the publications of Bell & Sejnowski 1995 (Infomax)
@@ -85,6 +85,8 @@ def infomax(data, weights=None, l_rate=None, block=None, w_change=1e-12,
         The maximum number of iterations. Defaults to 200.
     verbose : bool, str, int, or None
         If not None, override default verbose level (see mne.verbose).
+    force_random_seed : bool
+        If True, set random seed to a fixed value in each iteration.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


kazemakase:
Because you can use it to pass a different random number generator to the function, set a manual seed, and it seems to be somewhat standard in MNE (I want to keep the difference as small as possible).

cle1109:
But this seed is not used if `force_random_seed=True`, right? I'm just thinking if we could combine these two parameters into one to make it easier for the user.

jdammers:
@cle1109  @kazemakase @dengemann
according to the original implementation of Infomax (matlab version) the randon state is not fixed. The variability of multiple ICA decompositions like in ICASSO cannot be tested in this way.

kazemakase:
@jdammers @dengemann
With the current patch it is now possible to (a) use a seed that depends on the iteration step (like before), (b) use a fixed seed by specifying `random_state`, or (c) not touch the random state at all.

@cle1109 @dengemann
I did not want to touch the `random_state` parameter because I do not know how this parameter is supposed to behave. Furthermore, what if you want to specify a specific RNG *and* force the seed (well, maybe that's too utopic)?

dengemann:
Yes, this is relevant, please make a PR to MNE!

kazemakase:
> Yes, this is relevant, please make a PR to MNE!

In that case I suggest to continue this discussion at a new PR to MNE.

Once it's done, we can simply merge the upstream version into SCoT.

dengemann:
It's most important to have a solud reimplementation of the algo, even if this sacrifices reproducible results. Maybe the fixed random seed is rather a bug. I probably was biased by fastica at that time.

dengemann:
Api wise it should be done via random_state maybe redefine what None will do or allow for a flag 'iterative' or sth like that

dengemann:
Yes

cle1109:
@dengemann this is a good suggestion, that way we only need one parameter. @kazemakase at least for SCoT we could drop the function `check_random_state()` completely and just use the default random generator.

kazemakase:
@cle1109 
I guess we could do that... might be less work than inserting the `check_random_state` function again.

cle1109:
@kazemakase +1 from my side. Depending on the outcome of the discussion over at MNE, we can always pull another version from there later on. For now, I suggest that after removing `check_random_state()` and provided that all tests and Travis succeed, we merge this PR.

cle1109:
Thanks @kazemakase - now let's see if we can close #35 as well.

PR #60, opened 2015-08-04
Montagewarp

kazemakase:
To interpolate the EEG with spherical splines (topo plots, continous Laplacian) the electrodes are assumed to lie on a spherical surface. For most heads this is not the case :) To solve this discrepancy, arbitrary montages need to be transformed to a sphere. I found this to be not as trivial as it sounds:

- Simply projecting the locations to a shpere centered on the head causes *very* large distortions.
- It is better first scale the locations to be approximately spherical. However, this is still not ideal because it minimizes distortions in the scaled space, which is not the minimal distortion in the original location space.

I implemented the following improved procedure:

1. Fit an ellipsoid to the electrode locations (least squares).
2. Move electrodes to the closest point on the ellipsoid's surface (minimal distortion between actual positions and model).
3. Scale electrode locations by ellipsoid radii to get locations on the unit sphere.

cc @cle1109 @dengemann
What do you think about the procedure? Maybe have a quick look at the code? :)


kazemakase:
I forgot to mention the assumptions/limitations of the procedure.

The head ellipsoid is assumed to be axis oriented. This significantly reduces the degrees of freedom for fitting the ellipsoid. I don't think it is a limitation because many (all?) vendors ship axis aligned montage layouts.

If the layout is not axis aligned (e.g. individually measured electrode locations) it could be rotated (manually using landmarks, PCA) first. However this is a very custom and usage-dependent step, so I think it should not be implemented here but left to the user.

scot/eegtopo/geo_spherical.py
```diff
@@ -41,6 +41,11 @@ def vector(self):
         """position in 3d space"""
         return self._pos3d
 
+    @property
```
cle1109: Why is this a property?

scot/eegtopo/geo_spherical.py
```diff
@@ -41,6 +41,11 @@ def vector(self):
...
+    def list(self):
```
cle1109: `list` is a Python builtin type.

scot/eegtopo/warp_layout.py
```diff
@@ -0,0 +1,107 @@
...
+    p0 = locations - c  # original locations
+
+    l2 = 1 / np.sum(p0**2 / r**2, axis=1, keepdims=True)
+    p = p0 * np.sqrt(l2)  # initial approximation (projection of points towards center of ellipsoid)
+
+    fun = lambda x: np.sum((x.reshape(p0.shape) - p0)**2)              # minimize distance between new and old points
+    con = lambda x: np.sum(x.reshape(p0.shape)**2 / r**2, axis=1) - 1  # new points constrained to surface of ellipsoid
+    res = sp.optimize.minimize(fun, p, constraints={'type': 'eq', 'fun': con}, method='SLSQP')
+
+    return res['x'].reshape(p0.shape) + c
```
cle1109: Missing new line at EOF.

tests/test_eegtopo/test_warp.py
```diff
@@ -0,0 +1,53 @@
...
+
+        scale = np.random.rand(3) * 10 + 10
+        displace = np.random.randn(3) * 100
+        noise = np.random.randn(len(eeglocs), 3) * 5
+
+        assert_allclose(warp_locations(eeglocs), eeglocs, atol=1e-10)
+        assert_allclose(warp_locations(eeglocs * scale), eeglocs, atol=1e-10)
+        assert_allclose(warp_locations(eeglocs * scale + displace), eeglocs, atol=1e-10)
+        warp = warp_locations(eeglocs * scale + displace + noise)
+        assert_allclose(np.sum(warp**2, axis=1), 1, atol=1e-12)  # all locations on unit shpere
```
cle1109: Missing newline.

cle1109:
@kazemakase, I made some minor comments to your code, which looks very good. However, I didn't look at the math :-), this would take quite some time.

Regarding the warping process, can you quantify how much "better" this new algorithm is compared to a simple least squares fit to a sphere?

Also, I have a very basic question: how do you usually obtain the electrode locations that are then warped to a sphere? Are these vendor specifications? Or do you construct them (how)?

The [NFT toolbox](http://sccn.ucsd.edu/wiki/NFT) seems to do something similar. However, they warp the template head model (in our case a sphere, but they use some sort of standard BEM/FEM model) to the electrode locations, and not the other way round. I don't know why exactly they do that, but they claim that this procedure gives better results. However, I guess your goal is to end up with a sphere for simple visualization, right?

scot/eegtopo/geo_spherical.py
kazemakase: I'm afraid the honest answer is because this is the first Python code I ever wrote. Thought it would be a good idea to make this a property.

Improving the legacy code is a task for a separate PR. Unfortunately a few unrelated commits in seaked in here.

scot/eegtopo/geo_spherical.py
kazemakase: Yes.. should probably be named to_list(), and declared a function, not a property.
This is legacy code and should be dealt with in a different PR.

scot/eegtopo/warp_layout.py
kazemakase: Why does this always happen to me? :)

tests/test_eegtopo/test_warp.py
kazemakase: Strange. There is a newline in my local copy (which git claims is up-to-date).

cle1109: I only looked at the code in GitHub - maybe it doesn't show the newline and only complains explicitly if it is missing. So you should be fine!

kazemakase:
@cle1109 Thank you for your comments.

> Regarding the warping process, can you quantify how much "better" this new algorithm is compared to a simple least squares fit to a sphere?

I have not quantified the difference, but it should be easy to compare if required.

> Also, I have a very basic question: how do you usually obtain the electrode locations that are then warped to a sphere? Are these vendor specifications? Or do you construct them (how)?

I used to construct some of the default 10/10 positions directly on a sphere; there is no warping required. However, if you have vendor specifications the locations will be more accurate. These are typically not located on a sphere, and they often have locations that I don't know how to construct.
The reason I wrote this function is that I need to work with data in a rather strange layout that was exported from EEGLAB:
figure_1
(The image also shows qualitatively that an ellipsoid fits the locations better than a sphere.)

> I don't know why exactly they do that, but they claim that this procedure gives better results. However, I guess your goal is to end up with a sphere for simple visualization, right?

I guess they do that so the model fits the subject's head in size and shape, which makes their electrical simulations more realistic.
As you correctly guessed, my goal is to end with a sphere for topographic visualization. The sphere is also required for spherical spline interpolation and current density mapping.

cle1109:
> The reason I wrote this function is that I need to work with data in a rather strange layout that was exported from EEGLAB:

What are the blue and red dots in your figure?

tests/test_eegtopo/test_warp.py
kazemakase: Looks like this is a Windows issue. I checked out in Linux and there were no newlines.
(There was also no newline in the other file I already fixed)

kazemakase:
> What are the blue and red dots in your figure?

blue ... original locations (some above, some below the surface)
red ... displaced locations (all exactly on the ellipsoid surface)

cle1109:
I see. Now it would be interesting to see the spherical projection also for the non-ellipsoid variant.

kazemakase:
figure_2

> Head sphere center: [  4.35508369   0.24277626  53.89614775]
Head sphere radii: 99.6099303912
Minimum electrode displacement: 0.0673982553848
Average electrode displacement: 4.82613578338
Maximum electrode displacement: 14.6240988556

> Head ellipsoid center: [4.9194550920113569, 0, 41.589131838309996]
Head ellipsoid radii: [ 105.48640964   87.46552139  122.86889078]
Minimum electrode displacement: 0.0155670045262
Average electrode displacement: 1.73016102398
Maximum electrode displacement: 5.92991552219

Summary: Spherical mapping causes more than twice the electrode displacement on average.

cle1109:
Thanks @kazemakase, this looks really nice! +1 for merging.

PR #61, opened 2015-08-12
Testing with Python 3.4

kazemakase:
I've updated the testing environment to use Python version 3.4 instead of 3.3, and also updated the library dependencies (numpy, scipy, sklearn, matplotlib) to the newest versions available in Anaconda.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
     # This environment tests Python 2 support
     - DISTRIB="conda" PYTHON_VERSION="2.7" INSTALL_MKL="false"
       NUMPY_VERSION="1.9.2" SCIPY_VERSION="0.16.0"
-      SKLEARN_VERSION="0.16cond.1" MATPLOTLIB_VERSION="1.4.3"
+      SKLEARN_VERSION="0.16.1" MATPLOTLIB_VERSION="1.4.3"
       INSTALL_SCOT="true" RUN_EXAMPLES="false"
 
     # This environment tests Python 3 support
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,16 +7,16 @@ env:
   matrix:
     #- DISTRIB="ubuntu" PYTHON_VERSION="2.7" INSTALL_ATLAS="true"
 
-    # This environment tests the oldest supported anaconda env
+    # This environment tests Python 2 support
     - DISTRIB="conda" PYTHON_VERSION="2.7" INSTALL_MKL="false"
-      NUMPY_VERSION="1.8.1" SCIPY_VERSION="0.13.3"
-      SKLEARN_VERSION="0.14.1" MATPLOTLIB_VERSION="1.3.1"
+      NUMPY_VERSION="1.9.2" SCIPY_VERSION="0.16.0"
+      SKLEARN_VERSION="0.16cond.1" MATPLOTLIB_VERSION="1.4.3"
```
cle1109: The 2.7 build fails because it doesn't find this sklearn version - are you sure this is the correct version number?

cle1109:
Now that Python 3.5 is out, we should support it and set up testing environments.

kazemakase:
Although Anaconda supports Python 3.5, they don't have packaged numpy et al. for 3.5 yet.

Let's wait until our environment is easily available :)

cle1109:
Agreed. I expect this to happen very soon though (I am running a Git clone of NumPy myself because there's some weird bug that doesn't let me install it via `pip` :-)).

kazemakase:
Shall we merge this?

cle1109:
Yes, let's merge it and then work on Python 3.5 in another PR. We should also take care of that bug in the 2.7 env.

kazemakase:
I'll merge this. #75 will take care of the bug.

PR #62, opened 2015-08-17
Data (down)loader

mbillingr:
This PR deals with moving example data downloads from scot-data to scot. Previous discussion: scot-dev/scot-data#1.

Currently the PR only contains a first draft of the interface, which is open for discussion.

cc @cle1109

### DIFF
--- a/scot/exampledata.py
+++ b/scot/exampledata.py
@@ -2,7 +2,13 @@
 # http://opensource.org/licenses/MIT
 # Copyright (c) 2015 SCoT Development Team
 
-def load(dataset, datadir):
+from os.path import expanduser
+
+
+DEFAULT_STORAGE = expanduser("~/scot_data/")  # use config system instead
+
+
+def load(dataset, datadir=DEFAULT_STORAGE):
     """ Loads example dataset.
     
     If the requested dataset is not found in the location specified by
--- a/scot/exampledata.py
+++ b/scot/exampledata.py
@@ -1,0 +1,39 @@
+# Released under The MIT License (MIT)
+# http://opensource.org/licenses/MIT
+# Copyright (c) 2015 SCoT Development Team
+
+def load(dataset, datadir):
+    """ Loads example dataset.
+    
+    If the requested dataset is not found in the location specified by
+    `datadir` the function attempts to download it from the internet.
+    
+    Parameters
+    ----------
+    dataset : str
+        Which dataset to load. Currently only 'motor imagery' is supported.
+    datadir : str
+        Path to the storage location of example datasets. Datasets are downloaded
+        to this location if they cannot be found. If the directory does not exist
+        it is created.
+    Returns
+    -------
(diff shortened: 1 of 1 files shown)


kazemakase:
@agramfort We are currently discussing an automated downloader for example data. We don't want to inconvenience the user by putting downloads in arbitrary locations without them knowing. Do you have default locations for example data in MNE, or do you always ask the user where to put it?

agramfort:
in MNE we put datasets in examples folder for historical reasons but if
this folder is not readable we put in ~/mne_data. You can specify it with
an environ variable MNE_DATA

kazemakase:
Thank you @agramfort for your input.
Putting the data directly under the home folder seems to be the best solution so far... We could also make it a configurable default.

cle1109:
I kind of like the idea of having a `download` flag, which is `False` by default. The function interface could look like this:

    def load(dataset, download=False, datadir="~/scot_data"):

`download` could be `True`, `False`, and `"auto"` - the latter automatically downloads the file if it is not already in the specified location.

kazemakase:
I'm concerned with the use case of somebody trying out scot. What they likely want do is downloading the package and running an example.

If the default is not to donwload the example will not run, the user gets annoyed, and they use sift instead of scot.

cle1109:
Then set `download="auto"`? The question is then do we need this flag at all if you always want to download the file if it's not there? Probably not, then the only thing we have is the `datadir`. Sounds good to me.

kazemakase:
Agreed. shall we use a portable version of `~/scot_data` as the default?

cle1109:
Yes. Can you test if `~` is already portable? On Windows, this should point to the user's home directory `C:\Users\user`, so there is probably nothing to do there.

kazemakase:
this works: 

    def load(dataset, datadir=os.path.expanduser("~/scot_data")):

cle1109:
Cool - I guess we have our interface!

cle1109:
@mbillingr, what's the status of this PR? Are we ready to get rid of the `scot-data` package and move the example data into the main `scot` package?

mbillingr:
@cle1109 uh.. no.. I just realized we have only defined the interface, but it's totally lacking implementation.

Btw, I think the idea was to move only the loader into the `scot` package. The data itself should stay somewhere else so we don't carry a 50+MB data file around in the repository.

cle1109:
Only moving the loader is fine as long as we get rid of the scot-data package. Any ideas where we could put the binary? Does GitHub offer such an option?

cle1109:
Can we leave the data at https://github.com/scot-dev/scot-data/blob/master/scotdata/motorimagery.mat for now?

mbillingr:
I think we can leave the binary in the repository and let the downloader fetch it from there.

If at some point in the future GitHub complains that we create too much traffic we can always look for another storage location.

PR #65, opened 2015-08-18
Change order of data dimensions

mbillingr:
I'm sending the file assignment by email.


kazemakase:
We have to take care how `numpy.atleast_3d` behaves:

> (N,) --> (1, N, 1)
> (M, N) --> (M, N, 1)

This no longer fits the new order of dimensions. We want `(M, N) --> (1, M, N)` so that data without trials becomes data with one trials for functions that expect trials.

Should we implement our own `atleast_3d` function?

cle1109:
Hm, so you automatically convert non-epoched data to epoched (one trial) data in this case?

kazemakase:
Actually I only view the data as being epoched in one big epoch.

kazemakase:
The function `datatools.cat_trials` returns data of shape (samples*trials, channels).
Should we leave it like that? Then we can pass the result straight to sklearn functions such as Infomax or PCA.

Otherwise, if we change it to (channels, samples*trials), we'd have to transpose the result on every function call.

cle1109:
On the other hand, most scipy functions (e.g. detrend or lfilter) work along the last axis by default. But of course, in this case only the axis parameter must be set. I don't know, to me (samples, channels) makes more sense for 2D data (because that's what sklearn expects), but MNE uses (channels, samples). Can we introduce an axis argument?

cle1109:
And one more thing, what about compatibility? If we now change everything, existing scripts will not work. How should we handle this?

kazemakase:
> Can we introduce an axis argument?

Just transposing the output when required is probably easier.


> And one more thing, what about compatibility? If we now change everything, existing scripts will not work. How should we handle this?

We increase the version number and say the API changed :)

cle1109:
Should I also change `randomize_phase`?

kazemakase:
> Should I also change randomize_phase?

Yes. You can make it work along the last dimension, which makes it behave similar to scipy functions.

kazemakase:
How is this PR coming along?

cle1109:
Slowly. You really need to double-check my changes. Also, sometimes I feel
that just switching axes is not elegant since there might be some
unnecessary switching back and forth.

How is this PR coming along?

—
Reply to this email directly or view it on GitHub
<https://github.com/scot-dev/scot/pull/65#issuecomment-141974231>.

kazemakase:
Unfortunately there is always some switching back and forth, but it should only be required when interfacing with other libraries.

cle1109:
How do we handle our example data? Do we transpose it directly in the .mat file?

kazemakase:
I'm not sure... maybe it's better to leave the data untouched at the moment and adjust the loader.

cle1109:
OK, solved by https://github.com/scot-dev/scot-data/pull/3.

mbillingr:
Something is wrong with the Python 2.7 tests. Python 3 tests are passing. Have not checked the examples yet.

mbillingr:
The Yule-Walker equations are broken. They are not very important, so I just skipped the test :)

Examples work again, but we need to check if their results are consistent with the *master* branch.

cle1109:
Re Python 2.7, seems like the environment has not been set up correctly by Travis. It did work before, right? Usually, such errors disappear once we run the tests again (i.e. with the next commit).

mbillingr:
This happened twice now. I'll better check the logs in detail :(

mbillingr:
Doh! I guess the branch was too old. I've rebased to master. Let's see what happens...

mbillingr:
Finally! :relieved:

cle1109:
Nice, thanks for checking! So how do we proceed with this PR?

mbillingr:
I think some manual work is required:

1. We should do some manual testing to see if mundane things like data manipulation really work as expected.
2. We need to compare results from the examples against the master branch
3. Decide what to do about Yule-Walker equations

How did we say we wanted to transfer this huge API breaking change to master?

cle1109:
We didn't say anything about how we're going to introduce this change. Maybe we should just bump the version number to 0.2 and break the API. I guess that Scot is not such a huge package that we need to deprecate things in a graceful way (e.g. throwing a warning two versions before the break).

I can start with 1. since I'm analyzing new data anyway. Could you check some examples against master? Re 3, what do you suggest? What's the problem with YW?

cle1109:
Also, did I just break the tests?

mbillingr:
>  I guess that Scot is not such a huge package that we need to deprecate things in a graceful way

Exactly what I thought, too. Shall we bump the version number by a larger step in order to mark a large change?

> Re 3, what do you suggest? What's the problem with YW?

I don't remember why they were needed. I think the problem is with the arrangement of autocorrelation matrices, which needs to match the new data format.

> Also, did I just break the tests?

You did. the `-=`operator changes the input, which is not good :)

cle1109:
> Exactly what I thought, too. Shall we bump the version number by a larger step in order to mark a large change?

I don't know, 0.2 is a large step up considering that we took only tiny steps before. It's not that we add any functionality with this change, which is why I would not increase by too much.

 > You did. the -=operator changes the input, which is not good :)

Wait, what? I thought these two lines were equivalent? We did change the input before, didn't we?

mbillingr:
> I don't know, 0.2 is a large step up considering that we took only tiny steps before.

You mean like we are at 0.1.1 now? I guess you're right.

> Wait, what? I thought these two lines were equivalent? We did change the input before, didn't we?

Nah, if you reassign `x` it points to another object, but if you perform an operation on `x` you modify the object.
In other words, the `+` operator returns a new array, but the `+=` operator modifies the array in place.

cle1109:
> Nah, if you reassign `x` it points to another object, but if you perform an operation on `x` you modify the object. In other words, the `+` operator returns a new array, but the `+=` operator modifies the array in place.

Of course, but if you then assign the new array to the old name `x`, the result should be equivalent to the in-place operator. Right?

    x = x - np.mean(x, axis=1, keepdims=True)
    x -= np.mean(x, axis=1, keepdims=True)

mbillingr:
> Of course, but if you then assign the new array to the old name x, the result should be equivalent to the in-place operator. Right?

No. Try to see variables as C/C++ pointers. When you do `x = ...` you change where x points to. (have you seen the link in my previous post?)

cle1109:
> (have you seen the link in my previous post?)

Which link?

> No. Try to see variables as C/C++ pointers. When you do `x = ...` you change where x points to.

Well, yes, technically there's a difference. The in-place operator modifies the object in-place, whereas the assignment to the same name creates a new modified object. However, the resulting arrays should be equal (and by equality I don't mean the same object `is`, but the same value `==`).

So why does the in-place notation cause an error in the tests when the values of the results are the same?

mbillingr:
> So why does the in-place notation cause an error in the tests when the values of the results are the same?

The test is designed to make sure the function does not modify the input.
With the in-place operator you modify the input.

cle1109:
OK, now I get it :-). However, that's a pretty subtle behavior, don't you think?

mbillingr:
Yes it's subtle, but from a user perspective it sucks a lot if a function unexpectedly modifies your data. Just imagine you do `pca.fit(data)` and afterwards your data is all messed up...

Also, it's very pythonic not to modify the input :)

cle1109:
Of course, I agree that it is very Pythonic not to modify input parameters. I think my confusion was mainly caused by the fact that I thought the Numpy in-place operators are just syntactic sugar when in fact they're doing something else instead. That is, I didn't know they operated in-place just like `append` does for lists.

mbillingr:
I've rebased to master.

mbillingr:
The following examples are similar in this PR:
- [x] circular
- [ ] connectivity
- [ ] features
- [ ] MVARICA vs CSPVARICA  (big difference in MVARICA, smaller difference in CSPVARICA)
- [x] pca
- [ ] statistics
- [ ] timefrequency
- [ ] validation

Interesting that the circular example works while the others fail.

cle1109:
What does this list say? You checked examples that return identical results?

mbillingr:
Unfortunately, yes.

mbillingr:
Ok, I think I manged to salvage our commit history.

@cle1109 Please do not merge this branch into your local branch. And if you do, don't push it back here :)
You will have to replace your entire local branch with the new one. Best delete your transpose branch and check it out again...

mbillingr:
Looks good. All but the `features` example give similar results now.

I don't understand what's wrong with the features. Could you have a look at the example?

cle1109:
This is really weird. Strangely, the LDA weights are different. This is likely due to the features being slightly different - the difference is somewhere between 1e-14 to 1e-15 for each entry in the feature matrix. Do you have any idea where that difference might come from?

mbillingr:
Maybe the transposing process changed the order of some computations, which leads to these small numerical differences.

Are the differences really that small? It's surprising that they cause noticable changes in accuracy.

scot/datatools.py
```diff
@@ -153,6 +155,16 @@ def randomize_phase(data):
...
\ No newline at end of file
+    return np.fft.irfft(data_freq, data.shape[-1])
+
+
+def atleast_3d(x):
+    x = np.asarray(x)
+    if x.ndim >= 3:
+        return x
+    elif x.ndim == 2:
+        return x[np.newaxis, ...]
```
mbillingr: I don't understand this syntax with indexing `...`. How does it work?

mbillingr:
The changes (diff: master vs transpose) look plausible at a glance. I've added two small commits and commented on a line of code that I don't understand. Otherwise I'm fine with merging.

scot/datatools.py
cle1109: The ellipsis object `...` (don't ask me how this is valid Python syntax) automatically expands to all axes that have not explicitly specified. So, in our case this should be equivalent to `x[np.newaxis, :, :]`. Can you test this and maybe replace it (explicit is better than implicit)?

cle1109:
Feel free to either change the ellipsis or leave it in (I don't really have a preference here). Otherwise, +1 for merging!

cle1109:
Woo-hoo! Thanks @mbillingr!

Issue #40, opened 2014-12-17
Add sPDC method


dongqunxi:
Hi @kazemakase ,
I need to use the modified version of PDC, squared PDC, but I can not find it in the causal methods list. Can you give me some tips? Thanks!

agramfort:
@kazemakase your tugraz email address does not work. How can I reach you? github was my only option...

dongqunxi:
@kazemakase @agramfort Thanks for your helping on this issue. I edit the PDC function in https://gist.github.com/agramfort/9875439. And get the sPDC function, I am not sure if this modifying is opporiate, look forward to your suggestions, thanks!
```ruby
def sPDC(A, sigma=None, n_fft=None):
    """squared Partial directed coherence (sPDC)

    Parameters
    ----------
    A : ndarray, shape (p, N, N)
        The AR coefficients where N is the number of signals
        and p the order of the model.
    sigma : array, shape (N,)
        The noise for each time series.
    n_fft : int
        The length of the FFT.

    Returns
    -------
    P : ndarray, shape (n_fft, N, N)
        The estimated PDC.
    """
    p, N, N = A.shape

    if n_fft is None:
        n_fft = max(int(2 ** math.ceil(np.log2(p))), 512)

    H, freqs = spectral_density(A, n_fft)
    P = np.zeros((n_fft, N, N))

    if sigma is None:
        sigma = np.ones(N)

    for i in range(n_fft):
        B = H[i]
        B = np.abs(B) ** 2
        P[i] = B / np.sum(B, axis=0, keepdims=True)

    return P, freqs
```

kazemakase:
I think you forgot to invert H (i.e.`B = linalg.inv(H[i])`).
Otherwise it looks good to me.

dongqunxi:
@kazemakase Thanks a lot!

cle1109:
@kazemakase should we add this measure?

kazemakase:
Yes, I'd say so... that's why I left the issue open :)

cle1109:
I'd like to implement sPDC so that we can finally close this issue. I was wondering if the following implementation would do the trick (instead of the one suggested above, which looks much more complicated):

    @memoize
    def sPDC(self):
        """Squared partial directed coherence

        .. math:: \mathrm{sPDC}_{ij}(f) = \\frac{\left| A_{ij}(f) \right|^2}{\sqrt{A_{:j}'(f) A_{:j}(f)}}  # TODO: fix equation

        References
[...]

PR #67, opened 2015-08-18
Implemented squared PDC (sPDC)

cle1109:
Addresses #40.


cle1109:
IMO this could be merged now (provided that the test complete successfully).

kazemakase:
Is the sPDC really just the PDC? Why don't we just square the PDC? :)

cle1109:
Well yes, looking at the equation, I think that's what it is. Also, the name "squared PDC" could be a hint :-).

kazemakase:
:)

cle1109:
Do you think we can merge? This was almost too simple...

kazemakase:
Simple is good :) Let's merge it.

Issue #48, opened 2015-01-20
Set __version__


cle1109:
I think `__version__` is not set - I guess this should be set in `__init__.py` and read the version from the file `VERSION`.

kazemakase:
How do other libraries handle setting of __version__?

cle1109:
There are several options. I would go for option 4 [here](https://python-packaging-user-guide.readthedocs.org/en/latest/single_source_version.html#single-sourcing-the-version). That's how we do it in `setup.py` right now, so we could also read it in `scot/__init__.py`.

kazemakase:
I'm not sure if it is good practice to parse a text file when importing the library.

cle1109:
Not sure either - the caveat is of course that the file must be present. Alternatively, we could go with option 1, which involves setting `__version__` in `scot/__init__.py` and parsing this file in `setup.py` (which involves regexes).

kazemakase:
Can't we import scot in setup.py to get the version?

cle1109:
Sure, this should work. What do you think of #68?

cle1109:
Resolved by #68.

PR #68, opened 2015-08-19
Define __version__ in __init__ and read it in setup.py

cle1109:
Addresses #48.


kazemakase:
Looks good.  We should get rid of the VERSION file.

cle1109:
Feel free to merge.

Issue #69, opened 2015-08-19
Configuration


mbillingr:
Currently we have a scot/config.py which does not do much. We should use it to set (at least) the following options:

 - default backend
 - default example storage location

Furthermore, we should try to parse user/distribution specific configurations. Can we do it like this?

  1. try to load ~/.scot_config.py (or maybe ~/.scot/config.py)
  2. if that fails try /etc/scot_config.py (what's the windows equivalent?)
  3. if that faill too use scot/config.py

I'm not sure if it's possible to import python scripts from arbitrary locations. Otherwise we could use Json configuration.

cle1109:
Good idea. Why don't we just use a simple `config.ini`? There is a [configparser](https://docs.python.org/3/library/configparser.html) module which should handle this (although this might not work in Python 2 because it has been renamed - we would have to check that).

kazemakase:
We could use that. Should be easy to wrap for both Python versions:

    try:
        from configparser import ConfigParser
    except ImportError:
        from ConfigParser import ConfigParser

But I'm rather fond of the idea to use Python as the configuration language for Python :)
If it is possible, and unless there are security concerns, I'd vote for a configuration or startup script.

cle1109:
You're right, I didn't see that only the module name changed.

As much as I want to agree, I think a simple .ini file would be more than sufficient for our purposes and it is really a user-readable format. But this means we add in one more dependency (configparser, which is not that bad because it's in the standard library), so I'd be OK with having .py config files. Either way is fine with me.

Let's skip your second location (/etc corresponds to the registry on Windows and we don't want to mess with that - there are modules that deal with this sort of stuff; for example, PyQt has a nice configuration system that handles everything cross-platform). So I suggest that we

1. try to load ~/.scot_config.py
2. if the above fails, use the default configuration in scot/config.py.

kazemakase:
(I don't consider things from the standard library a dependency.)

Correct me if I'm wrong, but Python should be usereadable enough, especially if we only set a few variables. Also, I suppose that anyone who is willing to play with the configuration can be assumed to have some basic Python knowledge.

I'd like to use the .py config files if they can be loaded easily. Otherwise let's use the configparser.

At least on Linux I'd try to load the configuration from /etc. I imagine this could be a very useful features for distribution maintainers. For example, an Arch Linux package could include the example data and set the search path accordingly, without patching the source code.


[...]

PR #70, opened 2015-08-19
New configuration system

mbillingr:
Discussion: #69

This PR became rather big, because I had to change the backend mechanism which was closely tied to the old configuration system.

Currently two options are supported:

    [scot]
    verbose = True/False
    backend = builtin/sklearn

### DIFF
--- a/scot/__init__.py
+++ b/scot/__init__.py
@@ -1,6 +1,6 @@
 # Released under The MIT License (MIT)
 # http://opensource.org/licenses/MIT
-# Copyright (c) 2013 SCoT Development Team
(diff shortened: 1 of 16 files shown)


kazemakase:
Doh. Looks like the ConfigurationParser is not the same in Py2 and Py3.

cle1109:
Can you access the config with arguments in both 2 and 3 (as described [here](http://stackoverflow.com/questions/16407329/python-configparser-attributeerror-configparser-instance-has-no-attribute)?

mbillingr:
rebased to master

mbillingr:
@cle1109 Any objections to merging this?

Issue #73, opened 2015-09-18
Value of reducedim should be None instead of "no_pca"


cle1109:
When calling `mvarica` with the argument `reducedim="no_pca"`, I get an error in `pca`, because this function expects the argument to be `None` if no dimensions should be discarded.

I suggest we change the value of the `reducedim` argument in `mvarica` to `None` to be consistent with the `pca` function.

cle1109:
OK, actually the problem is just that line 81 incorrectly tested for `no pca` instead of `no_pca`. I'll fix it in a PR.

cle1109:
On a related note, is there a difference between performing no PCA vs. performing PCA and keeping all sources? This might impact our choice of arguments (and their values).

cle1109:
And I just noticed that the function `cspvarica` sets `reducedim=np.inf` if no dimensions are to be discarded. We should make all of these functions consistent.

kazemakase:
> On a related note, is there a difference between performing no PCA vs. performing PCA and keeping all sources? This might impact our choice of arguments (and their values).

Of course there is a difference between performing PCA and not performing PCA :) However, ICA should not care, *in theory*. In practice, there might be a difference. This option is mainly useful for comparing the method wwith the 'naiive' case of no PCA.

> And I just noticed that the function cspvarica sets reducedim=np.inf if no dimensions are to be discarded. We should make all of these functions consistent.

I'm not sure how consistent the functions can be. PCA-based dimensionality reduction can be specified in terms of variance reduction or number of components. For CSP only the number of components makes sense.

cle1109:
Hm. In both methods we can choose to keep all components, so we could have an argument that specifies exactly this in a consistent way for all related functions (`mvarica`, `cspvarica`, `pca`, `csp`, ...). Can't we say that `None` means we keep all dimensions? That way, I don't have to think about if I need to set the percentage of retained variance (1) or the number of components (`np.inf`)?

kazemakase:
`np.inf` could work for both functions, but I think `None` is more pythonic. It also feels sensible that setting `reducedim=None` should lead to no dimensionality reduction.

cle1109:
Fixed by #74. #77 addresses the consistency issue.

PR #74, opened 2015-09-18
Changed argument value to "no_pca" instead of "no pca"

cle1109:
Addresses #73.


kazemakase:
What's wrong with the test?

cle1109:
Looks like something's wrong with the testing env. We should take a look at the envs we use (and add one for Python 3.5).

kazemakase:
I rebased to the current master branch where the test/env bugs are fixed. *Fingers crossed*

PR #75, opened 2015-09-21
Migrating travis infrastructure

kazemakase:


kazemakase:
Weird... the library bug is gone, but now one of the test cases fails.

PR #78, opened 2015-09-21
Building the documentation on Windows

kazemakase:
Looks like I can get it to work.


kazemakase:
Do you know where the documentation gets the scot version number from?

cle1109:
No, I only ever built it, but never actually took a closer look at the sources...

kazemakase:
I just realized it's hardcoded :(

cle1109:
This sucks, but for the moments let's just keep it that way and hope we don't forget to update the version with every update of the docs.

kazemakase:
No problem, I fixed this already :p

cle1109:
:+1:

PR #82, opened 2015-12-21
Refactored test_var_sklearn

cle1109:
The tests work with automated tools such as `nose`.

Why do we need to test `self.assertTrue(np.all(x == y))`?

### DIFF
--- a/tests/test_var_sklearn.py
+++ b/tests/test_var_sklearn.py
@@ -5,7 +5,7 @@
 import unittest
 
 import numpy as np
-from sklearn import linear_model
+from sklearn.linear_model import Ridge, RidgeCV, Lasso, LassoLars, ElasticNet
 
 from scot.backend_sklearn import VAR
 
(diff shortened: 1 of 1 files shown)


mbillingr:
To check that the function does not modify the input.

I am not 100% sure if sklearn guarantees this on all functions, so it's probably best to leave it in.

cle1109:
OK, but then we need to put it into a test* method. What do you think of the current implementation?

mbillingr:
Is it necessary to put the assertions into `test*` methods?
Otherwise we could put them in the common `_fit` method, which would allow us to perform different checks.

cle1109:
I think I made the code simpler again by restricting each test to one method and putting the common stuff into `setUp`. What do you think?

I also removed the check that test if `sklearn` changes `X`. I think this really never happens, and if it does, it is a `sklearn` bug. We don't need to catch bugs of other packages here.

mbillingr:
> We don't need to catch bugs of other packages here.

That's true.


Your solution looks clean, but all the boilerplate code hurts my eyes :)

Alternatively, we could go object oriented: an abstract base class that implements the setup and tests, but not the fitting. The latter is performed in derived classes, one for each fitting routine.
Something like that: http://stackoverflow.com/a/14930907/3005167
Thus, we could centralize common code and get a nice separation between tests and sklearn functions.

cle1109:
Right, I agree that the current implementation is a bit clunky, but probably the most straightforward one. Plus, it has the advantage that it actually works as opposed to the previous implementation :-).

I would suggest that we merge this PR and open a new issue for the object-oriented refactoring of the test cases.

PR #85, opened 2015-12-23
Refactored tests

mbillingr:
Each sklearn model has it's own test class now. These classes inherit common tests and can define their own model specific tests if required.

@cle1109 what do you think?

### DIFF
--- a/tests/test_var_sklearn.py
+++ b/tests/test_var_sklearn.py
@@ -5,7 +5,7 @@
 import unittest
 
 from numpy.testing.utils import assert_array_almost_equal
-from numpy.testing.utils import assert_almost_equal
+from numpy.testing.utils import assert_equal
 
 import numpy as np
 from sklearn.linear_model import Ridge, RidgeCV, Lasso, LassoLars, ElasticNet
@@ -40,17 +40,17 @@ class CommonTests(unittest.TestCase):
 class TestRidge(CommonTests):
     def setUp(self):
         super(TestRidge, self).setUp()
-        self.var = VAR(10, Ridge(alpha=115))
+        self.var = VAR(10, Ridge(alpha=100))
 
 
 class TestRidgeCV(CommonTests):
     def setUp(self):
         super(TestRidgeCV, self).setUp()
-        self.var = VAR(10, RidgeCV(alphas=np.logspace(-3, 3, 20)))
+        self.var = VAR(10, RidgeCV(alphas=[10, 100, 1000]))
 
     def test_alpha(self):
         self.var.fit(self.x)
-        assert_almost_equal(self.var.fitting_model.alpha_, 115.0, decimal=-1)
+        assert_equal(self.var.fitting_model.alpha_, 100)
 
 
 class TestLasso(CommonTests):
--- a/tests/test_var_sklearn.py
+++ b/tests/test_var_sklearn.py
@@ -46,10 +46,16 @@ class TestRidge(CommonTests):
 class TestRidgeCV(CommonTests):
     def setUp(self):
         super(TestRidgeCV, self).setUp()
+        # Provide three candidates for alpha.
         self.var = VAR(10, RidgeCV(alphas=[10, 100, 1000]))
 
     def test_alpha(self):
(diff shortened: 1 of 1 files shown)


tests/test_var_sklearn.py
```diff
@@ -1,74 +1,71 @@
...
         self.var0.coef = np.array([[0.2, 0.1, 0.4, -0.1], [0.3, -0.2, 0.1, 0]])
         self.x = self.var0.simulate((1000, 100))
+        self.var = VAR(2)
 
     def tearDown(self):
         pass
 
     def test_fit(self):
-        var = VAR(2)
+        var = self.var
```
cle1109: Why do you create the name `var` here? Couldn't you just use `self.var.fit(self.x)` in the next line?

tests/test_var_sklearn.py
```diff
@@ -1,74 +1,71 @@
...
-    def test_fit_elasticnet(self):
-        var = VAR(10, ElasticNet(alpha=0.01, l1_ratio=0.5))
-        var.fit(self.x)
-        b0 = np.zeros((2, 20))
-        b0[:, 0:2] = self.var0.coef[:, 0:2]
-        b0[:, 10:12] = self.var0.coef[:, 2:4]
-        self.assertTrue(np.all(np.abs(b0 - var.coef) < 0.02))
+    def test_alpha(self):
+        self.var.fit(self.x)
+        assert_almost_equal(self.var.fitting_model.alpha_, 115.0, decimal=-1)
```
cle1109: What does this method test exactly? What is `115.0` here?

tests/test_var_sklearn.py
mbillingr: It was easier to copy and paste the old code :)

Also, below, `var` is often used multiple times per line. Replacing it with `self.var` would make the code less readable.

I am not sure if this is good practice, though. Should we replace `var` with `self.var` throughout this function?

tests/test_var_sklearn.py
```diff
@@ -1,74 +1,71 @@
...
-    def test_fit_elasticnet(self):
-        var = VAR(10, ElasticNet(alpha=0.01, l1_ratio=0.5))
-        var.fit(self.x)
-        b0 = np.zeros((2, 20))
-        b0[:, 0:2] = self.var0.coef[:, 0:2]
-        b0[:, 10:12] = self.var0.coef[:, 2:4]
-        self.assertTrue(np.all(np.abs(b0 - var.coef) < 0.02))
+    def test_alpha(self):
+        self.var.fit(self.x)
+        assert_equal(self.var.fitting_model.alpha_, 100)
```
mbillingr: @cle1109 
> What does this method test exactly? What is 115.0 here?

Is it clearer now?

tests/test_var_sklearn.py
cle1109: I don't know, I think creating an alias here is not really necessary because `self.var` is not that hard to read, and if you also create a local `var` then we have too many names with `var` floating around. So I'd stick to `self.var`.

tests/test_var_sklearn.py
cle1109: Yes, it's clearer now. But why do you test if `alpha == 100` when you specify three `alpha` values?

mbillingr: Because RidgeCV is supposed to select the best alpha (using cross-validation), which is 100.

cle1109: OK, we're approaching the answer :-) - why is the best alpha 100? Where do you set that?

mbillingr: Ah, yes.. the dreaded question :p

It is not set. I just happen to know that the best alpha is close to 100 for this particular testing data.

cle1109:
@mbillingr do we want to merge this before the transpose PR?

mbillingr:
Let's do it afterwards for the same reason as #30.

mbillingr:
rebased to master

mbillingr:
@cle1109 can we merge this?

PR #86, opened 2016-01-26
Check and update Travis CI

mbillingr:
Does Travis have problems?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,16 +9,22 @@ env:
   matrix:
     #- DISTRIB="ubuntu" PYTHON_VERSION="2.7" INSTALL_ATLAS="true"
 
+    # This environment tests compatibility with oldest supported package versions
+    - DISTRIB="conda" PYTHON_VERSION="2.7" INSTALL_MKL="false"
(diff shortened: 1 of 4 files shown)


mbillingr:
Seems to be working, here...

mbillingr:
@cle1109 
With this PR tests are now performed against the newest versions of numpy, scipy, etc.

I've also added a test for older version of these libraries, so we can specify minimum requirements.

cle1109:
Great - feel free to merge. BTW, what are our minimum requirements now? We could mention this in our docs.

mbillingr:
It works with numpy 1.8.0, scipy 0.13.3, scikit-learn 0.15.0, and matplotlib 1.4.0 on Python 2.7.

We could probably go lower than that, but I think two releases behind current is good enough for the moment. I believe the gap will grow in the future ;)

cle1109:
Very good, this sounds reasonable. Could you add these requirements somewhere? At least it should be added to `README.md`, and maybe adding it to the online doc would also be a good idea.

mbillingr:
Feel free to merge if you're happy.

PR #89, opened 2016-01-27
de-randomized examples

mbillingr:
All examples now seed the numpy random number generator, which makes results repeatable.


mbillingr:
@cle1109 Somehow the comment was lost when squashing the commits. I've added the newline.

cle1109:
LGTM

Issue #96, opened 2016-02-02
Improve test coverage


mbillingr:
Not much to say. These are the worst covered files:

file | coverage | lines missed
--- | --- | ---
scot/varica.py | 71% | **26**
scot/plainica.py | 76 % | 6
scot/var.py | 78% | **31**
scot/backend_builtin.py | 85% | 3
scot/pca.py | 87% | 6
scot/varbase.py | 89% | **16**
scot/ooapi.py | 94% | **18**
scot/plotting.py | 96% | **10**

cle1109:
I'm closing this for now. If we feel that we need to significantly increase our coverage again, let's just open a new issue/PR.

PR #99, opened 2016-02-02
Improve Coverage

mbillingr:
Addresses #96

### DIFF
--- a/scot/parallel.py
+++ b/scot/parallel.py
@@ -18,7 +18,10 @@ def parallel_loop(func, n_jobs=1, verbose=1):
     try:
         from joblib import Parallel, delayed
     except ImportError:
-        n_jobs = None
+        try:
+            from sklearn.externals.joblib import Parallel, delayed
+        except ImportError:
+            n_jobs = None
 
     if n_jobs is None:
         if verbose >= 10:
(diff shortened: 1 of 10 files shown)


cle1109:
LGTM. I wonder why the coverage increase isn't shown here.

mbillingr:
It is shown when you click on "Show all checks".

cle1109:
Any idea why the bisection test fails?

Also, since we'll be treating the issues with parallelization in #30, maybe you should also remove the corresponding tests for now? Or is this not related?

mbillingr:
This is not related. The problem is here: 

```
    if trform(b) >= maxdelta and verbose:
        print('Bisection: could not find initial interval.')
        print(' ********* Delta set to zero! ************ ')
        return 0
```

The test checks if the function returns 0 because `b` is too high. This works locally because I have set `verbose`. On travis verbose is `False` so it does not return 0.

cle1109:
LGTM. We're almost at 90%. Can you add a test to cover the remaining 0.035%? This would be great from a psychological perspective :-).

mbillingr:
I've meant this as a sort of open ended PR, anyway :)

cle1109:
We need to merge it at some point - otherwise the increased coverage remains in this branch.That's why I think 90% would be a good threshold to merge this (and continue increasing coverage in another PR later).

mbillingr:
Right.

mbillingr:
When writing these tests I came across a bug in CSPVARICA (we forgot to fix data orientation in one place).

Hooray for testing!

cle1109:
Nice! I think we can merge this now, right?

PR #100, opened 2016-02-03
Removed in-module execution of unit tests

mbillingr:
Tests are invoked by tools such as nosetets, not by the test modules themselves.

Invokation of `unittest.main()` was removed from all test scripts.


cle1109:
Nice and necessary :-) - +1 for merge!

PR #104, opened 2016-02-03
[Hotfix] New config broke examples

### LINKED ISSUE #103: New config system broke some examples
For example, `circular.py` does not work anymore. The problem is in `__init__` of `Workspace`. `self.var_ = self.backend_['var'](**var)` returns a `dict`.

mbillingr:
Actually it broke default backend assignment in `Workspace`.

Should fix #103.


cle1109:
LGTM - if all tests and examples work now, +1 for merging.

PR #106, opened 2016-02-04
Improve coding style

### LINKED ISSUE #79: Coding style
We should try to adhere to a consistent coding style. I suggest we stick to what everyone else does, namely PEP8 and NumPy docstrings, such as scikit-learn (that's where I mainly look up style-related things if in doubt).

If you agree, I will create a PR after we're done with #65, the big one.

cbrnr:
Fixes #79. @mbillingr can you confirm if it is OK to move the `cat_trials` import outside of the function?

### DIFF
--- a/scot/__init__.py
+++ b/scot/__init__.py
@@ -1,9 +1,8 @@
 # Released under The MIT License (MIT)
 # http://opensource.org/licenses/MIT
-# Copyright (c) 2013-2015 SCoT Development Team
+# Copyright (c) 2013-2016 SCoT Development Team
 
-""" SCoT: The Source Connectivity Toolbox
-"""
+"""SCoT: The Source Connectivity Toolbox."""
 
 from __future__ import absolute_import
 
@@ -13,7 +12,7 @@ config = config.load_configuration()
 from .backendmanager import BackendManager
 backend = BackendManager()
 
-# register backends that come shipped with scot
+# register backends shipped with SCoT
 from . import backend_builtin
 from . import backend_sklearn
 backend.activate(config.get('scot', 'backend'))
--- a/scot/backend_builtin.py
+++ b/scot/backend_builtin.py
@@ -5,17 +5,15 @@
 """Use internally implemented functions as backend."""
 
 from __future__ import absolute_import
+import scipy as sp
 
 from . import backend
+from . import datatools, pca, csp
+from .var import VAR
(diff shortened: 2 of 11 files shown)


mbillingr:
Yes, that is OK. In fact, you could also move the `dot_special` import outside. 

In principle, we could move all local imports and the scipy import outside. The local imports are available anyway, and scipy is a obligatory dependency of scot. Only `sklearn` is optional and should remain inside the function.

cle1109:
@mbillingr I've changed the input argument order in some functions. Could you please check if this is OK, or are we unnecessarily breaking the API? I thought the new order makes more sense.

cle1109:
Hmmmm. Seems like rebasing didn't work as expected. Again.

mbillingr:
Did you pull/push in pycharm? :p

I think, what I did was to reset `HEAD` to the last good commit. I think in this case the commit is `Fix tests` (f7a3f7f)

Note that after doing a rebase your local branch and the branch at origin will have diverged. You *must not* merge this diverged branches. Instead use `push --force` to overwrite the old commits at origin.

cle1109:
OK, that did it. I should do all git-related stuff on the command line. Although I'm sure there's an option in PyCharm that I need to check in order to perform a rebase correctly.

I think I'm going to revert the changed argument order. The statistics example produces different results.

mbillingr:
The problem with changing the argument order is that you will never know for sure if you found all the places where the function is used.

cle1109:
Agreed. That's why I reverted it.

mbillingr:
@cle1109 can you rebase this branch to the current master?

cle1109:
Looks like rebasing in PyCharm worked this time - I had to enable Force Push in the settings.

mbillingr:
Force push is also what you have to do after a rebase in the console.

But I still don't trust PyCharm :)

cle1109:
Yes, it still feels like it could do all sorts of weird stuff in the background, but so far it seems to be working. Most of the time, I'm just too lazy to open a terminal to rebase and force push. And BTW, PyCharm has a menu entry for "rebase my GitHub fork" which has worked for me so far.

cbrnr:
@mbillingr could you please check if the changes are OK? If yes, please merge it - I want to avoid blowing up this PR more, because this means a lot of manual merging when rebasing. We can continue to improve the code style in another PR.

Issue #107, opened 2016-02-05
Travis setups require MKL?


cle1109:
Two of our three Travis setups fail because some MKL-related libs are not found. However, we've explicitly specified `INSTALL_MKL="false"` in all of our setups. It only works in the first setup (Python 2.7, latest packages).

mbillingr:
Why did it sop working, now?

cle1109:
I have no idea. But it is very annoying since we don't have our tests.

PR #108, opened 2016-02-05
What's wrong with travis MKL?

mbillingr:
see #107

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 sudo: false
 
 virtualenv:
-  system_site_packages: false
+  system_site_packages: true
   
 env:
   matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ env:
 
     # This environment tests Python 3 support
     - DISTRIB="conda" PYTHON_VERSION="3.4" INSTALL_MKL="false"
-      NUMPY_VERSION="1.10.1" SCIPY_VERSION="0.16.0"
+      NUMPY_VERSION="1.10.2" SCIPY_VERSION="0.16.1"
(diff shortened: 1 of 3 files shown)


cle1109:
Do you know why there are 2 Travis build jobs (pr and push)?

mbillingr:
I do not know, but I guess that is because the branch exists in the main repository. So it starts one build because I pushed to a branch in this repo, and another build because it's included in a pull request.

mbillingr:
Didn't change anything. I hope this is just a passing issue with the anaconda packages ...

cle1109:
Yep. It's not the first time that Travis is having strange hiccups. Is this worth opening an issue at https://github.com/travis-ci/travis-ci?

mbillingr:
I don't think the problem is related to travis. Seems more like an anaconda problem.

cle1109:
Can you try to commit something and see if it works now?

cle1109:
Only the first env is working - is it because Fortran is installed? Strange that we get MKL-related errors even though we explicitly uninstall all MKL-related Anaconda packages.

mbillingr:
I have no idea :)

mbillingr:
Although we explicitly uninstall mkl, it is reinstalled again...

mbillingr:
The installation **with** MKL works now.

The installation **without** MKL does not work.
- Although we do not explicitly request it, MKL is installed along with other packages
- Although MKL is installed, it does not appear to be used (there is no message about the trial period).
- Although MKL is not used, we get the error: `Intel MKL FATAL ERROR: Cannot load libmkl_avx.so or libmkl_def.so.`

@cle1109 Any ideas?

PR #110, opened 2016-02-08
Unify _construct_var_eqns

### LINKED ISSUE #94: Unify construction of VAR equations
Currently, the functions `_construct_var_eqns` in varbase.py and `_construct_var_eqns_rls` in var.py share a lot of code.

I propose to construct the equations in one place.

There are at least two options:
 1. to have one function that can optionally create equations with and without RLS.
 2. let `_construct_var_eqns_rls` call `_construct_var_eqns` to construct the equations and then pack the RLS part on top.

cle1109:
Fixes #94. @mbillingr, what do you think?

### DIFF
--- a/scot/var.py
+++ b/scot/var.py
@@ -207,7 +207,7 @@ class VAR(VARBase):
     def _construct_eqns_rls(self, data):
         """Construct VAR equation system with RLS constraint.
         """
-        return _construct_var_eqns_rls(data, self.p, self.delta)
+        return _construct_var_eqns(data, self.p, self.delta)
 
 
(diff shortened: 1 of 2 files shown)


mbillingr:
LGTM

cle1109:
If you're happy go ahead and merge it.

mbillingr:
I wish the tests would run properly...

cle1109:
Yes, me too. At least the first env passes, which is a good sign. I think it's safe to merge this PR, but we should try to get Travis up and running ASAP. Unfortunately, I have no idea what the problem could be. Did you check the config in other repos (such as mne-python or scikit-learn)?

PR #112, opened 2016-02-08
Fix MKL Error in Travis

### LINKED ISSUE #107: Travis setups require MKL?

mbillingr:
Closes #108
Closes #107

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,13 +17,13 @@ env:
 
     # This environment tests Python 2 support
     - DISTRIB="conda" PYTHON_VERSION="2.7" INSTALL_MKL="false"
-      NUMPY_VERSION="1.10.1" SCIPY_VERSION="0.16.0"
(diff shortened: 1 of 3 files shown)


cle1109:
Why is it working now?

mbillingr:
I've increased numpy/scipy versions :)

cle1109:
What? That's good news, but I wonder if we should create test environments that always fetch the latest package versions of Numpy, Scipy, scikit-learn, and matplotlib instead of relying on fixed version numbers that will stop working eventually... Anyway, thanks for fixing this one!

mbillingr:
Normally it's the other way round - fixed versions are more stable because they are, well, fixed :)

Maybe I did't pay attention during the last upgrade, or maybe there is a hidden dependency that caused trouble.

Issue #97, opened 2016-02-02
Fix ICA warnings in tests


mbillingr:
Some tests cause these warnings:
- overflow in Infomax
- FastICA does not converge

Let's try to fix them.

mbillingr:
We may be able to improve FastICA convergence by changing initialization weights (see #59).

cle1109:
+1 for changing init weights to the identity matrix.

cle1109:
cf #114

cle1109:
Thanks @mbillingr, this means we only have to fix the FastICA convergence issue.

PR #114, opened 2016-02-08
Fixed overflow in Infomax

mbillingr:
The overflow warning we saw in infomax was caused by testing data that had values across several orders of magnitude. 

Since this is not realistic for EEG anyway I simply reduced the magnitude of the data to make the warning go away...


cle1109:
Nice. Did we fix the other warning related to FastICA too?

mbillingr:
Did you? I didn't?

cle1109:
Nope. But please go ahead and merge this one - it fixes half of #97.

PR #115, opened 2016-02-09
Anaconda 2.5

mbillingr:
Anaconda 2.5 comes by default with MKL. It can be disabled by installing `nomkl`.
Details [here](https://www.continuum.io/blog/developer-blog/anaconda-25-release-now-mkl-optimizations).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
       INSTALL_SCOT="true" RUN_EXAMPLES="false"
 
     # This environment tests Python 2 support
-    - DISTRIB="conda" PYTHON_VERSION="2.7" USE_MKL="false"
+    - DISTRIB="conda" PYTHON_VERSION="2.7" USE_MKL="true"
       NUMPY_VERSION="1.10.4" SCIPY_VERSION="0.17.0"
       SKLEARN_VERSION="0.17.0" MATPLOTLIB_VERSION="1.5.1"
       INSTALL_SCOT="true" RUN_EXAMPLES="false"
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,19 +14,21 @@ env:
       NUMPY_VERSION="1.8.0" SCIPY_VERSION="0.13.3"
       SKLEARN_VERSION="0.15.0" MATPLOTLIB_VERSION="1.4.0"
       INSTALL_SCOT="true" RUN_EXAMPLES="false"
+      COVERAGE="false"
 
     # This environment tests Python 2 support
     - DISTRIB="conda" PYTHON_VERSION="2.7" USE_MKL="true"
       NUMPY_VERSION="1.10.4" SCIPY_VERSION="0.17.0"
       SKLEARN_VERSION="0.17.0" MATPLOTLIB_VERSION="1.5.1"
       INSTALL_SCOT="true" RUN_EXAMPLES="false"
+      COVERAGE="false"
 
     # This environment tests Python 3 support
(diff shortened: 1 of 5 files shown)


mbillingr:
Damn, I think I branched from the wrong branch :)

cle1109:
Good that it's not only me who messes up the repo :-P.

mbillingr:
Damn again! :)

mbillingr:
Yay! With a bit of practice it becomes possible to fix all these GIT mess-ups.

cle1109:
Sure - rewriting history rocks!

Back on topic, so this means we're using MKL only with Python 3? Shouldn't we also use MKL with Python 2? This way, we would test the two default Anaconda envs. In addition, we have an "oldest supported" env, which could be the non-MKL one with Python 2 and old packages.

cle1109:
So isn't Travis going to run the tests?

mbillingr:
Checks passed. They ran in the background.

Logs look good too - I've enabled MKL in the Python 3 test and disabled it in the Python 2 and the compaitibility tests.

This is what we get:

Default numpy build with MKL
```
    package                    |            build
    ---------------------------|-----------------
    libgfortran-1.0            |                0         170 KB
    mkl-11.3.1                 |                0       121.2 MB
    python-3.4.4               |                0        15.3 MB
    numpy-1.10.4               |           py34_0         5.9 MB
    ...
```

Special nomkl numpy build that's using openblas instead
```
    package                    |            build
    ---------------------------|-----------------
    libgfortran-1.0            |                0         170 KB
    nomkl-1.0                  |                0          402 B
    openblas-0.2.14            |                3         3.5 MB
    python-2.7.11              |                0        12.0 MB
    numpy-1.10.4               |     py27_nomkl_0         6.0 MB
    ...
```

cle1109:
OK! Do you want to use the default current Python 2.7 version (i.e. with MKL)?

mbillingr:
Hmm... I don't know.

MKL and BLAS are supposed to be equivalent and it's not our task to verify that. So, do we need to test both, with and without MKL?
Do we want the Python 2 and Python 3 tests to be identical?
Do we want to avoid too many test runs?

It may be best to stick to the default which everybody may be using in the future (and which hopefully also makes our tests faster). This would mean using MKL in the tests, but we could disable it in the compatibility check. What do you think?

cle1109:
Exactly, we shouldn't even have to deal with the question MKL or non-MKL. Therefore, let's just take the standard Anaconda installs - which means, starting with 2.5, MKL-enabled 3.5 and 2.7. We can even use the default MKL 2.7 with the oldest supported packages.

What do you mean with disabling MKL in the compatibility check?

mbillingr:
> What do you mean with disabling MKL in the compatibility check?

I meant not to use MKL with the old packages.

cle1109:
Ah - yes, this sounds good. Could you update the envs accordingly?

Plus, the three envs should set the same variables - currently, the Python 3 env doesn't set the `RUN_EXAMPLES` variable, and so on.

Finally, we should update the Python 3 env to Python 3.5.

mbillingr:
Anything else?

cle1109:
Since you're asking, can you increase the verbosity of the tests? Currently, there's only one line of dots...

mbillingr:
Since you mention it, I've been planning to do that for a long time :)

cle1109:
:-)

mbillingr:
Btw, feel free to merge.

PR #117, opened 2016-02-09
Add (back) badges

cle1109:
@mbillingr if you like it go ahead and merge.


mbillingr:
I don't like it. I love it! :D

PR #122, opened 2016-02-10
Update doc template

cle1109:
I've simplified the website structure and switched to a different template.

### DIFF
--- a/doc/source/_templates/layout.html
+++ b/doc/source/_templates/layout.html
@@ -1,14 +1,5 @@
 {% extends "!layout.html" %}
 
-
-{% block rootrellink %}
-        <li><a href="{{ pathto('index') }}">home</a>|&nbsp;</li>
-        <li><a href="{{ pathto('examples/index') }}">examples</a>|&nbsp;</li>
-        <li><a href="{{ pathto('gallery') }}">gallery</a>|&nbsp;</li>
-        <li><a href="{{ pathto('search') }}">search</a>|&nbsp;</li>
-{% endblock %}
-
-
 {% block relbar1 %}
 
 <div style="background-color: white; text-align: left; padding: 10px 10px 15px 15px">
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -125,19 +125,19 @@ mpl_example_sections = (('misc', 'Miscellaneous Examples'),
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+html_theme = 'classic'
 
 # Theme options are theme-specific and customize the look and feel of a theme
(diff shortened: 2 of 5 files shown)


cle1109:
D'oh! I didn't branch from master...

cle1109:
@mbillingr what do you think? It's definitely better than what we currently have, so I suggest that we use this template for now.

mbillingr:
I don't know what to think, because I don't know what it looks like now :)

How do you navigate to the gallery in the new layout?

cle1109:
Can you pull my changes to your local repo and build the HTML pages? Navigation is not much different than before, I just removed the sidebar and the four menu points because I think they were confusing.

mbillingr:
Looks okay. Two issues, though:

1. Example plots are missing. This may be a problem of my Windows build, but please verify if you have them before uploading.
2. The Gallery became inaccessible, because it was only linked from the top menu.

cle1109:
1. Examples plot work fine on my system.
2. Right - do you know how you created the gallery? Is this a Sphinx plugin? How is it generated? How was the menu entry generated? I would like to add a link under "About SCoT" right after the examples.

mbillingr:
1. OK

2. The gallery is created by a Sphinx extension. I think it works by parsing the directory where the examples put their images, creates a thumbnail for each of them and links to the example's page. The menu entry was not generated, I simply added the gallery entry to `_templates/layout.html` and it worked :) I guess we will have to find out where the resulting html is put and link to it in the "about" section.

cle1109:
I've added the gallery. However, I'm not really happy with the structure of the website. For example, the gallery doesn't seem to belong to the "About SCoT" section, even though I've included it in the corresponding file. I wanted to change the title "SCoT examples" to "Examples", but this change is reverted every time I build the docs.

I think a major overhaul of the website is necessary. Or rather, I was wondering why we need a separate website for our docs - couldn't we just put them right here on our GitHub site? We do have a Wiki which we could use for docs that go beyond the main `README.md` (and I think we need to expand `README.md` to include all necessary information).

mbillingr:
We can't put an automatically generated API reference into the Wiki. I also think it's nice to have an example gallery. Otherwise I'm perfectly comfortable with using the Wiki. But if we need to have docs for reference and gallery we can put the documentation in the docs as well.

cle1109:
You're right, the API reference is nice to have. Then let's try to work on the website structure. In the meantime, do you think we can use the modified website from this PR?

mbillingr:
Sure. As far I understand, the structure is still the same, only the navigation bars (top and side) are removed. right?

cle1109:
Exactly. For now, this should be OK. For the future, we need to think about the structure in more detail.

mbillingr:
Then let's merge this PR and work out a nice structure for the future. Maybe we can learn from other projects.

PR #133, opened 2016-02-15
Cross validation tests with scikit-learn

### LINKED ISSUE #132: Support scikit-learn cross-validation objects
Currently, SCoT uses CV implemented in `xvschema.py`. It would be nice if we supported scikit-learn CV objects as well.

mbillingr:
I've added some tests that compare our and sklearn's cross validation routines. They also use sklearn cross-validation in VAR optimization. However, this requires wrapping of the function, because our functions take a `skipstep` argument that does not exist in scikit-learn.

Resolves #132

### DIFF
--- a/tests/test_xvschema.py
+++ b/tests/test_xvschema.py
@@ -5,11 +5,12 @@
 import unittest
 
 import numpy as np
+from numpy.testing import assert_array_equal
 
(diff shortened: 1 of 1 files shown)


cle1109:
LGTM. Feel free to merge!

PR #134, opened 2016-02-16
Move tests inside scot

### LINKED ISSUE #80: Move tests inside scot package
People recommend to have the `tests` package as a subpackage inside the actual package. Therefore, I suggest that we move our `tests` package to `scot/tests`.

cle1109:
Fixes #80.

Anything else we need to adapt? Instead of `cd`ing into the `tests` folder, we could stay in the root and call all programs from there. But if this works, we can also leave it as is.


cle1109:
Wait, what? Why did coverage increase by 3.8%?

mbillingr:
This build lists coverage of 51 files. Other builds only have 29 files.

Among these 51 files are the `test_xyz.py` test scripts. I guess coverage now tracks all the Python files under the `scot` directory. Is there a way to prevent that?

cle1109:
Which files inside scot don't we want to cover?

mbillingr:
Everything inside `scot/tests`.

cle1109:
Interesting. How can the tests be covered by tests? We should find out how other projects with tests inside the package run their coverage.

mbillingr:
The tests are actually 100% covered, because every line is executed.

cle1109:
Ah. I thought that somehow coverage is computed from the number of lines executed after running all tests. I didn't know that running the test files themselves also count.

cle1109:
Maybe this is even what we want: http://stackoverflow.com/questions/1628996/is-it-possible-exclude-test-directories-from-coverage-py-reports

@agramfort, do you include the tests in your coverage?

cle1109:
@mbillingr, I propose to go ahead and include our tests in the coverage report. Scikit-learn seems to include the tests as well (or at least, they don't explicitly exclude them): https://github.com/scikit-learn/scikit-learn/blob/master/.coveragerc

If this file is really the place to configure Coveralls, should we go ahead and add this configuration too? It might be a good idea to exclude certain things such as `setup.py`, `external`, and `doc` (but not the tests).

cle1109:
Any idea why `test_eeglocations` fails (1) now and (2) for Python 3.5 only?

mbillingr:
Maybe this is a random fail? I've restarted the travis build to find out. Nevertheless, we should examine this issue in more detail.

cle1109:
By random fail you mean related to a "bad" random seed? Do we have to seed this test as well?

I didn't know that you can manually restart Travis builds, nice!

mbillingr:
This is strange. The test sets a random seed, so it should not arbitrarily fail.

There is a restart button at the top right of the travis build page:

unbenannt

cle1109:
Now it passes. Huh?

mbillingr:
Yep.

cle1109:
Up to you - pretend the failed Travis build never happened and merge, or try to investigate what the problem was? We should be able to safely merge this despite of the failed test, what do you think?

mbillingr:
Let's merge this and wait if it happens again. Maybe it was just a travis hickup.

cle1109:
What about creating a `.coveragerc` file?

mbillingr:
Why? What does it do?

cle1109:
I guess it tells Coveralls which files to ignore.

mbillingr:
Ah, right. Good idea. I'm not entirely sure which files to ignore (setup.py and external/ would be good start).

cle1109:
Maybe it's best to specify the `sources` argument.

mbillingr:
If we specify to include the whole `scot` directory, will it also include `external`?

cle1109:
I'm not in front of my PC anymore, but I guess you could try it with the standard coverage.py on your local machine.

cle1109:
Hm. I think coverage is collected by `nosetests` via a coverage plugin, which likely uses `coverage`. I guess since everything is working, and others also include their test scripts in the coverage report, we shouldn't change anything.

mbillingr:
Yes, but it would be nice if we could exclude the external dependencies.

cle1109:
I think the `nosetests` plugin also reads `.coveragerc`. We could create this file with the following contents:

```
[run]
branch = True
source = scot
include = */scot/*
omit = 
    */setup.py
    */scot/external/*
```

I think this should cover everything in the scot folder (recursively), but exclude the things we specify in `omit`. I just don't know why we need the `source = scot` line then.

mbillingr:
Maybe it's redundant to specify a starting point. What does `include = */scot/*` do?

cle1109:
> What does `include = */scot/*` do?

No idea - I've pushed the config to see what happens.

mbillingr:
Looks like it did not work. `external/infomax_.py` is still covered.

(`setup.py` was never covered, even before. I guess that's because it is not in the scot/ directory.)

cle1109:
BTW, I was wondering why you use `xvfb-run --server-args="-screen 0 1024x768x24"` to execute programs in the Travis scripts...

mbillingr:
Matplotlib needs an X-server.

cle1109:
And the tests produce figures? Does this make sense?

cle1109:
Looks good now! The only thing I don't understand is why Coveralls includes the tests, whereas `nosetests` doesn't (if you check the Travis build log for the Python 3.5 env). Any ideas?

In any case, I checked scikit-learn, and there it's the exact same thing. I guess we should be good to go.

mbillingr:
OK

Issue #124, opened 2016-02-11
Add __print__ (or __repr__) to Workspace class


cle1109:
It would be nice if printing a Workspace object showed some basic information. I'm never sure if `__print__` or `__repr__` should be used for this purpose though.

mbillingr:
I think `__repr__` and `__str__` should be used for this purpose.

For the summary `__str__` is better, because `__repr__` should be a representation of the object that can be used to reconstruct it by evaluating the string.

cle1109:
Ah, I didn't see that. Currently, you need to `print(ws)` in order to see this info. However, it would be great if this was also shown when just typing the object in the interactive interpreter, i.e. `ws`. I assume this would work if we move the implementation of `__str__` to `__repr__`?

mbillingr:
Right. Ipython seems to show `__repr__` and not `__str__`. We could change that (I think `__str__` falls back to `__repr__`when unimplemented, so it's enough to implement one of those), but let's first find out if it is good practice to show a summary in `__repr__`.

cle1109:
Yes, it is. It seems like `__repr__` should always be implemented. Printing an object falls back to `__repr__`. If you need a clearer textual representation, use `__print__` to create a modified version of the info in `__repr__`. Bottom line: yes, let's move our `__print__` implementation to `__repr__`.

mbillingr:
Does this answer the question if it is good or bad practice to show only a summary in `__repr__` without fully representing the instance?

(I can't check stackoverflow at the moment, because I'm in a hotel and for some reason only https:// sites work.)

cle1109:
More or less. https://stackoverflow.com/questions/1436703/difference-between-str-and-repr-in-python

mbillingr:
I take the as essintial point that `eval(repr(ws)) == ws` holds. This will be difficult in our case because the workspace contains EEG and source data.

Do you think it's ok to violate this principle?

In this case we could create a `__repr__` that shows the information in a somewhat compressed form (i.e. one liner), and a `__str__` that shows a nicely formatted version of the same information. Or we use only `__repr__` and to hell with standards :)

cle1109:
In Python, we're most often talking about conventions and not standards - and that's a good thing :-).

It makes sense to go with what everybody does. I can't decide right now, because I need to dig deeper into this topic (I don't particularly like the SO post I linked, it was just the first one I found). Let's try to figure this out before we decide how to proceed.

mbillingr:
[...]

PR #138, opened 2016-02-16
Remove return values

### LINKED ISSUE #137: Do we need return values for do_mvarica and do_cspvarica?
I was wondering if we really need return values for the functions `do_mvarica` and `do_cspvarica`. Somehow, I expect these functions to store the computed values in the `Workspace` object (which they do) - so why do they need to return the results as well?

cle1109:
Resolves #137

### DIFF
--- a/scot/ooapi.py
+++ b/scot/ooapi.py
@@ -207,11 +207,6 @@ class Workspace(object):
             'class' fits a different model for each class, and
             'trial' fits a different model for each individual trial.
 
-        Returns
-        -------
-        result : class
-            see :func:`mvarica` for a description of the return value.
-
         Raises
         ------
         RuntimeError
@@ -232,7 +227,6 @@ class Workspace(object):
         self.activations_ = dot_special(self.unmixing_.T, self.data_)
         self.mixmaps_ = []
         self.unmixmaps_ = []
-        return result
 
     def do_cspvarica(self, varfit='ensemble'):
         """ Perform CSPVARICA
@@ -246,11 +240,6 @@ class Workspace(object):
             'ensemble' (default) fits one model to the whole data set,
             'class' fits a different model for each class, and
             'trial' fits a different model for each individual trial.
-
-        Returns
-        -------
-        result : class
-            see :func:`cspvarica` for a description of the return value.
 
         Raises
         ------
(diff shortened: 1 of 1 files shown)


mbillingr:
Would it make sense to return `self` for operator chaining?

Btw, 'addresses' is no magic keyword :)

cle1109:
> Would it make sense to return `self` for operator chaining?

Hm. It's pretty unlikely that someone would want to chain ICA, MVARICA, and CSPVARICA. But if it doesn't hurt - why not? Does anything get printed in an interactive session if we return `self`? I'd prefer if no output was produced.

> Btw, 'addresses' is no magic keyword :)

They should really add it.

mbillingr:
> Hm. It's pretty unlikely that someone would want to chain ICA, MVARICA, and CSPVARICA

hehe, yes :)
But this would make sense: `ws.set_data(data).do_mvariaca().get_connectivity('PDC')`

> Does anything get printed in an interactive session if we return self? I'd prefer if no output was produced.

Uhm.. yes. I think it prints the return value :(

>> Btw, 'addresses' is no magic keyword :)

> They should really add it.

+1

cle1109:
> But this would make sense: `ws.set_data(data).do_mvariaca().get_connectivity('PDC')`

Right, that's a pretty awesome use case. We should return self. Even though it prints the return value. Which is `Workspace.__str__` or `Workspace.__repr__`?

mbillingr:
> Right, that's a pretty awesome use case. We should return self.

Do the other methods (such as `set_data`) return `self`? Maybe we should add it there too :)

> Which is Workspace.__str__ or Workspace.__repr__?

In IPython:
```
In [7]: class A:
   ...:     def __repr__(self):
   ...:         return('repr')
   ...:     def __str__(self):
   ...:         return('str')
   ...: 

In [8]: A()
Out[8]: repr
```

cle1109:
> Do the other methods (such as `set_data`) return self? Maybe we should add it there too :)

+1. Where do you want to return `self`?

Yes! Since we decided in #124 that `__repr__` returns just a one-liner, this does not interfere too much with IPython output.

mbillingr:
> Where do you want to return self?

`set_locations`, `set_premixing`, `set_data`, `set_used_labels`, `do_mvarica`, `do_cspvarica`, `do_ica`, `remove_sources`, `fit_var`, `optimize_var`

cle1109:
All methods done. Can you also check if the docstring on the return value looks right?

mbillingr:
More accuractely, the docstring would say "reference to the Workspace object".

But this is Python, and references are the default, so I don't think we have to be explicit here. If you agree feel free to merge.

PR #140, opened 2016-02-19
Add option to change clipping in topos (and set default to 'electrodes')

cle1109:
This would be a useful little addition IMO.


cle1109:
@mbillingr anything else that needs to be addressed?

mbillingr:
LGTM, feel free to merge.

Issue #141, opened 2016-02-22
Add workspace method to keep sources


cle1109:
In addition to `ws.remove_sources`, we should have a `ws.keep_sources`.

cle1109:
In addition, would it make sense not to manipulate the matrices directly, but to have a mask that marks the sources that should be used in all computations?

mbillingr:
Yes, this makes sense.

I think we would need getter methods for the matrices to relieve the user from dealing with the masks.

cle1109:
Do we? All functions working with the matrices could automatically apply the mask.

mbillingr:
If users want to get the matrices they should not have to apply the mask themselves.

cle1109:
But what if they want the full non-masked matrices?

mbillingr:
Nobody forces them to remove components, first ;)
And they can access the class attributes, directly too.

cle1109:
OK. Then we need a method to reset the mask, e.g. `ws.reset_components`. I assume we will only modify that for the `Workspace` class?

Which variables do you want to decorate with `@property`? How should we modify the names? This could get cumbersome if we call the property `self.mixing_` and the corresponding private variable `self._mixing_` - don't you think?

mbillingr:
I had been thinking in simpler terms of implementing `get_mixing()` and `get_unmixing()` methods.

cle1109:
That's so Java and not Pythonic :-). So you think we need to consider `mixing` and `unmixing`. What about `activations?` Anything else?

mbillingr:
That's not Java, that's C++ :)
I can't think of anything else at the moment.

cle1109:
Well, why don't we use the decorators then?

mbillingr:
I never said we shouldn't use `@property`. Properties are rather nice and we can use them, here. (And they are simpler to use than I thought.)

Nevertheless, I am surprised that properties are supported in Python. I always thought that one of the principles of being Pythonic is *explicit is better than implicit*. With a property the user thinks they just access a value stored in the class, but actually it involves a function call where the value is computed on the fly. A getter function would be much more explicit, and therefore much more Pythonic, no?

cle1109:
> I never said we shouldn't use @property. Properties are rather nice and we can use them, here. (And they are simpler to use than I thought.)

OK, then let's go for it.

[...]

PR #145, opened 2016-02-25
Add keep_sources method

cle1109:
Addresses at least part of #141 (I didn't implement a mask yet).

### DIFF
--- a/scot/ooapi.py
+++ b/scot/ooapi.py
@@ -368,6 +368,15 @@ class Workspace(object):
         self.unmixmaps_ = []
         return self
 
+    def keep_sources(self, keep):
+        """Keep only the specified sources in the decomposition.
+        """
+        if self.unmixing_ is None or self.mixing_ is None:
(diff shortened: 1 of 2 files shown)


mbillingr:
Don't forget to add a test for this method.

Do you plan to add the mask in this PR too?

cle1109:
I think we should add the mask in another PR.

Also offtopic: many tests need artificial test data (such as the tests I just wrote) - we should create a central facility that generates data for all tests. Once we move our real example data out of `scot-data`, we could then even write (optional) tests that work on the example data. What do you think?

mbillingr:
I agree. In the beginning I intended to have such a central test data generation facility. There is a `generate_testdata.py` that does not do very much at the moment.

It would be awesome if we could get a realistic leadfield matrix from a head model to generate testing data.

cle1109:
OK. I'll open a new issue.

Also, go ahead and merge this PR if you're happy with it.

PR #147, opened 2016-03-02
Add random_state

### LINKED ISSUE #139: Add option to seed random number generator
In #89, we added a random seed to our examples. I think it would be nice if we had an option to set the seed globally, e.g. `ws.set_seed` for the OO API.

cbrnr:
Fixes #139.

So far, MVARICA supports the `random_state` argument. I'll add the other functions soon. In the meantime, please let me know if my changes are OK @mbillingr.

### DIFF
--- a/scot/csp.py
+++ b/scot/csp.py
@@ -5,7 +5,7 @@
 """Common spatial patterns (CSP) implementation."""
 
 import numpy as np
-from scipy.linalg import eig
+from scipy.linalg import eigh
     
 
 def csp(x, cl, numcomp=None):
@@ -62,8 +62,8 @@ def csp(x, cl, numcomp=None):
         sigma2 += np.cov(x2[t, :, :]) / x2.shape[0]
     sigma2 /= sigma2.trace()
 
-    e, w = eig(sigma1, sigma1 + sigma2, overwrite_a=True, overwrite_b=True,
-               check_finite=False)
+    e, w = eigh(sigma1, sigma1 + sigma2, overwrite_a=True, overwrite_b=True,
+                check_finite=False)
 
     order = np.argsort(e)[::-1]
     w = w[:, order]
--- a/scot/datatools.py
+++ b/scot/datatools.py
@@ -10,6 +10,8 @@ Tools for basic data manipulation.
 
 import numpy as np
 
+from .utils import check_random_state
+
 
 def cut_segments(x2d, tr, start, stop):
     """Cut continuous signal into segments.
@@ -115,7 +117,7 @@ def dot_special(x2d, x3d):
                            for i in range(x3d.shape[0])])
 
 
-def randomize_phase(data):
+def randomize_phase(data, random_state=None):
     """Phase randomization.
 
     This function randomizes the spectral phase of the input data along the
@@ -154,9 +156,10 @@ def randomize_phase(data):
         plot(y.T + [1.5, -1.5]), axis([0,1000,-3,3])
         plt.show()
     """
+    rng = check_random_state(random_state)
     data = np.asarray(data)
     data_freq = np.fft.rfft(data)
-    data_freq = np.abs(data_freq) * np.exp(1j*np.random.random_sample(data_freq.shape)*2*np.pi)
+    data_freq = np.abs(data_freq) * np.exp(1j*rng.random_sample(data_freq.shape)*2*np.pi)
     return np.fft.irfft(data_freq, data.shape[-1])
 
 
--- a/scot/datatools.py
+++ b/scot/datatools.py
@@ -13,6 +13,16 @@ import numpy as np
 from .utils import check_random_state
 
 
+def atleast_3d(x):
+    x = np.asarray(x)
+    if x.ndim >= 3:
+        return x
+    elif x.ndim == 2:
+        return x[np.newaxis, ...]
+    else:
+        return x[np.newaxis, np.newaxis, :]
+
+
 def cut_segments(x2d, tr, start, stop):
     """Cut continuous signal into segments.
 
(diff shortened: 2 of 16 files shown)


cle1109:
I don't understand the coverage report:

1. How come the report is shown _before_ the Travis build for Python 3.5 is finished? I thought this build triggers Coveralls?
2. Why did coverage increase? What is the reference/baseline? The previous commit? The current master?

cle1109:
@mbillingr can you take a look at https://github.com/cle1109/scot/commit/c83eb5b4f05445a8ea018a79c4973948355658c5 - I've disabled two tests in this function because I have no idea what they do. Also the last line in the test raised an error related to the eigenvalue decomposition (see Travis log in the previous commit) - I think not only is the input data linearly dependent(?), but there is also no `assert` statement to catch any errors.

cle1109:
Either I'm completely blind or there really is a problem with `conda`. Since I've tried long enough to resolve this, I reported the issue at https://github.com/ContinuumIO/anaconda-issues/issues/683.

cle1109:
I guess that's all - do you see anything I forgot?

cle1109:
Furthermore, what's up with Travis?

scot/csp.py
```diff
@@ -59,7 +59,7 @@ def csp(x, cl, numcomp=None):
         sigma2 += np.cov(x2[t, :, :]) / x2.shape[0]
     sigma2 /= sigma2.trace()
 
-    e, w = eig(sigma1, sigma1 + sigma2, overwrite_a=True, overwrite_b=True, check_finite=False)
+    e, w = eigh(sigma1, sigma1 + sigma2, overwrite_a=True, overwrite_b=True, check_finite=False)
```
mbillingr: How is this change related to the random state?

cle1109: It's not :-). But when I stumbled over it, I thought it had to be changed.

scot/datatools.py
```diff
@@ -154,17 +168,48 @@ def randomize_phase(data):
...
     """
+    rng = check_random_state(random_state)
     data = np.asarray(data)
     data_freq = np.fft.rfft(data)
-    data_freq = np.abs(data_freq) * np.exp(1j*np.random.random_sample(data_freq.shape)*2*np.pi)
+    data_freq = np.abs(data_freq) * np.exp(1j*rng.random_sample(data_freq.shape)*2*np.pi)
     return np.fft.irfft(data_freq, data.shape[-1])
 
 
-def atleast_3d(x):
```
mbillingr: Why did you remove this function - and why are things still working after removing it?

cle1109: I didn't remove it, I just moved it to the top (which isn't necessary of course, but in the course of debugging the cyclic import I tried moving it around).

scot/tests/test_ooapi.py
```diff
@@ -238,6 +238,25 @@ def test_plotting(self):
...
+        api.do_mvarica(random_state=1)
+        mixing2 = api.mixing_
+        assert_allclose(mixing1, mixing2)
+
+        # test CSPVARICA
+        api.do_cspvarica(random_state=1)
+        mixing1 = api.mixing_
+        api.do_cspvarica(random_state=1)
+        mixing2 = api.mixing_
+        assert_allclose(mixing1, mixing2)
```
mbillingr: Why do you use `assert_allclose` instead of `assert_allequal`? Since we put the same numbers in the algorithms, there should be no numerical differences...

mbillingr:
What about travis?
There are some changes in this PR that I think do not quite fit the topic. Moving functions between `utils` and `datatools` is a good idea, but not really related to the random state?

cle1109:
> What about travis?

It's working now. I guess it was related to a `conda` update, which rendered our `numpy` version non-functional. Upgrading `numpy` to 1.8.2 solved the problem.

However, I don't understand why Travis doesn't trigger Coveralls anymore. Any ideas?

> There are some changes in this PR that I think do not quite fit the topic. Moving functions between `utils` and `datatools` is a good idea, but not really related to the random state?

Yes, it was related. Since I added `check_random_state` to `utils` and `acm` required `atleast_3d` from `datatools`, this resulted in a cyclic import. Moving `acm` solved the issue, but of course I could have also moved the `import` to the bottom of the file (which is ugly).

scot/tests/test_ooapi.py
cle1109: You're right, I'll replace it with `assert_allequal`.

scot/csp.py
mbillingr: Hehe, ok, but now we have to pollute this PR by discussing why this is necessary.

Why is it necessary? :)

scot/datatools.py
mbillingr: Oh, I missed that.

scot/csp.py
cle1109: Because we're dealing with Hermitian matrices. If our matrices are not Hermitian, this indicates a problem with our covariance matrices (e.g. due to a lack of data and/or no regularization). In such cases, `eigh` fails and tells us that something is wrong. `eig` on the other hand silently produces complex eigenvalues, which we usually ignored. For example, there was a problem with one test case, which generated a sine as input signal.

mbillingr:
> However, I don't understand why Travis doesn't trigger Coveralls anymore. Any ideas?

The builds show up on the coveralls page, so I guess the problem is related to coveralls, not to travis. Otherwise I have no idea, yet.

> cyclic import.

I understand.

mbillingr:
Btw, 
> Why did coverage increase? What is the reference/baseline?

The baseline is the current master. At the bottom of the report (coveralls page) it links to the baseline commit.

cbrnr:
@mbillingr can you check if we've caught all randomness? And if so, please feel free to merge.

mbillingr:
Can you rebase?

cbrnr:
Yup. Done.

PR #149, opened 2016-03-04
Test conda environment

cle1109:
Upgraded to numpy 1.8.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,9 @@ virtualenv:
   
 env:
   matrix:
-    #- DISTRIB="ubuntu" PYTHON_VERSION="2.7" INSTALL_ATLAS="true"
-
     # This environment tests compatibility with oldest supported package versions
     - DISTRIB="conda" PYTHON_VERSION="2.7" USE_MKL="false"
-      NUMPY_VERSION="1.8.0" SCIPY_VERSION="0.13.3"
+      NUMPY_VERSION="1.8.2" SCIPY_VERSION="0.13.3"
       SKLEARN_VERSION="0.15.0" MATPLOTLIB_VERSION="1.4.0"
       INSTALL_SCOT="true" RUN_EXAMPLES="false"
       COVERAGE="false"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,38 +1,41 @@
-language: python
+language:
+  python
 
-sudo: false
+sudo:
+  false
 
 virtualenv:
-  system_site_packages: false
+  system_site_packages:
+    false
   
 env:
   matrix:
-    # This environment tests compatibility with oldest supported package versions
+    # oldest supported package versions on Python 2
     - DISTRIB="conda" PYTHON_VERSION="2.7" USE_MKL="false"
       NUMPY_VERSION="1.8.2" SCIPY_VERSION="0.13.3"
       SKLEARN_VERSION="0.15.0" MATPLOTLIB_VERSION="1.4.0"
       INSTALL_SCOT="true" RUN_EXAMPLES="false"
       COVERAGE="false"
 
-    # This environment tests Python 2 support
+    # Python 2
     - DISTRIB="conda" PYTHON_VERSION="2.7" USE_MKL="true"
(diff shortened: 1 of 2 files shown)


cle1109:
@mbillingr, do you have an idea how I can show the output of the scripts `install.sh` and `run_tests.sh`? I would like to see the output of the `conda` command, and I don't understand why it is not in the Travis log.

cle1109:
Somehow the conda problem magically disappeared. @mbillingr I'm going to merge this because it affects all our builds. I hope this is OK.

mbillingr:
> @mbillingr, do you have an idea how I can show the output of the scripts install.sh and run_tests.sh? I would like to see the output of the conda command, and I don't understand why it is not in the Travis log.

They are shrunk in the log. If you click on the line the script output is expanded.


What was the conda problem? Are you sure it was caused and fixed by the numpy version?

cle1109:
> They are shrunk in the log. If you click on the line the script output is expanded.

Thanks - I didn't see this tiny light gray rectangle :-).

> What was the conda problem? Are you sure it was caused and fixed by the numpy version?

I guess so. It seemed to involve a bit of magic as well, because when I first tested with `numpy` 1.8.2, it didn't work either. On the next day, it worked.

mbillingr:
> On the next day, it worked.

Have you tried reverting back to 1.8.0 on the next day?

cle1109:
I'm so happy that it finally worked that I don't dare to touch it anymore! Should I?

mbillingr:
In principle, I think we should change the minimum environment as little as possible (only if we add features to SCoT, that are incompatible with older packages).

In this case, I don't really care if we depend on numpy 1.8.0 or 1.8.2 (although 1.8.0 slightly nicer because this means we can claim to support the whole 1.8.x branch). Anyway, if you want to leave it please update the readme.

cle1109:
Yep, I agree - we should not change anything unless absolutely necessary (and in addition, we should really try to lower the versions even more, because I think they might still work).

Let's see if 1.8.0 works: #150

cle1109:
No, it doesn't. So it really was related to a change in how `conda` packages `numpy.`

mbillingr:
Strange. We had this problem before. I solved it by explicitly installing fortran after conda was prepared. Then I changed it to install everything in one go. When switching to the mkl/nomkl feature I removed explicit installation of `libgfortran`, which worked until recently.

I don't know why this keeps breaking. Does the anaconda team change old packages?

cle1109:
Yes, I think they're changing old packages.https://github.com/ContinuumIO/anaconda-issues/issues/683

mbillingr:
Does not sound to me like they were changing the old packages themselves, but the libgfortran package.

How do we proceed?

We can stay with numpy 1.8.2, but do we know it won't break again with another update?
We can follow the adivce and support only the newest 1.8.x release.
Finally, we can try to "pin" the libgfortran package - but only for the compatibility environment. I'm not sure how difficult this is.

cle1109:
Yes, I think they changed the version of the `libgfortran` package. However, I don't really understand the details of their packaging. For example, there is no separate `libgfortran` package available in `conda`. I suggest that we keep 1.8.2 for now, but we should be prepared for these conda-related problems in the future.

mbillingr:
> I don't really understand the details of their packaging. For example, there is no separate libgfortran package available in conda

Actually there is one: https://anaconda.org/anaconda/libgfortran, but it is only available/required for Linux. I guess you did `conda search` on mac/windows, where it does not show up.
There are two versions of this package: 1.0 and 3.0. The latter is 4 days old, which coincides with the start of our problems.

> I suggest that we keep 1.8.2 for now, but we should be prepared for these conda-related problems in the future.

OK. Can you fix the numpy version in `readme.md`?

cle1109:
> Actually there is one: https://anaconda.org/anaconda/libgfortran, but it is only available/required for Linux. I guess you did `conda search` on mac/windows, where it does not show up.

True.

> There are two versions of this package: 1.0 and 3.0. The latter is 4 days old, which coincides with the start of our problems.

OK, that makes sense. Still, I think it is pretty annoying that Anaconda renders older versions of otherwise working packages unusable (at least on Linux).

PR #152, opened 2016-03-09
Set reducedim to None by default in Workspace

### LINKED ISSUE #148: cspvarica does not support relative (percent) reducedim
I noticed that in contrast to `mvarica` (which uses PCA), `cspvarica` does not support a relative number for `reducedim` (e.g. 0.99 to retain 99% of the variance). This is a problem because the `Workspace` sets the default value of `reducedim` to 0.99.

What do you suggest to do @mbillingr? I guess it is not possible to run `cspvarica` with a relative number for `reducedim?`

cle1109:
Fixes #148. Feel free to merge @mbillingr.

### DIFF
--- a/scot/ooapi.py
+++ b/scot/ooapi.py
@@ -63,7 +63,7 @@ class Workspace(object):
     `plot_f_range` : (int, int)
         Lower and upper frequency limits for plotting. Defaults to [0, fs/2].
     """
-    def __init__(self, var, locations=None, reducedim=0.99, nfft=512, fs=2, backend=None):
+    def __init__(self, var, locations=None, reducedim=None, nfft=512, fs=2, backend=None):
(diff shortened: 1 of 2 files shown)


mbillingr:
I think the test that fails relies on the old default reduction. We may have to set the value manually.

mbillingr:
This time an unrelated test failed: #153.
I'll restart travis too see if it was a random problem.

cle1109:
Interesting. It fails only on Python 3.5 (which works fine on my local machine). I'm losing faith in Travis...

mbillingr:
I don't think it's travis' fault. It seems more like a random problem to me.

cle1109:
It works now. However, it would be good to fix this behavior in #153. Feel free to merge this one.

PR #155, opened 2016-03-15
Added codecov support

### LINKED ISSUE #154: Coveralls
Any idea why we no longer get coverage reports from Coveralls?

mbillingr:
Fixes #154 

Let's see if it works that easily :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,3 +39,4 @@ script:
 
 after_success:
   - if [[ "$COVERAGE" == "true" ]]; then coveralls; fi
+  - if [[ "$COVERAGE" == "true" ]]; then codecov; fi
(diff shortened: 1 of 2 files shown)


cbrnr:
Nice! +1 for merging!

mbillingr:
Codecov reports a slightly lower coverage than Coveralls. I think that is because coveralls has something like "partial hits" that are not counted. I don't know yet what they signify :)

cbrnr:
Since all coverage services are using `coverage`, I'm pretty sure they can be set to yield exactly identical results. However, I'm happy with the Codecov number - everything >90% is great!

cbrnr:
Why didn't @codecov-io comment after the last commit?

mbillingr:
Maybe it only updates the first comment to avoid clutter?

cbrnr:
It didn't update the first comment, or did it?

mbillingr:
Coverage did not change in this PR. It updated the comment in #156, though.

PR #156, opened 2016-03-15
Removed some tests to try coverage

mbillingr:
We must not merge this PR :)


mbillingr:
Comments work, but I don't understand why we do not get build status checks.

cbrnr:
Does the first comment get updated to reflect the current coverage?

mbillingr:
> Does the first comment get updated to reflect the current coverage?

Yes. It even says so in the the last line of the comment :) (at least it claims that it does)

cbrnr:
OK :-)

Re build status checks - GitHub doesn't know we're using Codecov, because Travis triggers it directly. Maybe we need to add a hook so that GitHub knows we want Codecov coverage also as a status check?

mbillingr:
As I understood, Codecov should tell GitHub about the status. It has access through my account:

image

cbrnr:
Checks seem to work now - did the codecov config file solve it?

Also, why are there 2 checks (codecov/patch and codecov/project)?

mbillingr:
Yes, it started to work after adding the config.

`codecov/project` checks the total coverage of the project.
`codecov/patch` checks only the coverage of lines modified by the PR.

This information helps to keep coverage high. For example, consider we added a new function with 10 lines. If the function is not tested, total coverage would decrease by less than a percent. Nobody would probably care about this small drop. However, the patch coverage would report this PR with 0% coverage because all new lines are untested. This should be immediately obvious. 

I'll add a commit to test this.

cbrnr:
Does it make sense? Both checks seem to report the correct coverage. But what are these commit hashes they refer to? I can't find 9613d8f or 25d816a.

cbrnr:
@mbillingr can you try disabling suggestions in `codecov.yml` line 22?

mbillingr:
> But what are these commit hashes they refer to? I can't find 9613d8f or 25d816a

I have been wondering about that too.

mbillingr:
Hmm... 25d816a is the last commit in the current master, but I don't know about 9613d8f

mbillingr:
I think 9613d8f is a "virtual" merge commit. It has two parents: 25d816a and 74f60a6

These are the current master and the last commit in this PR.

Why Codecov reports on this strange commit eludes me...

cbrnr:
I wonder why they use this virtual merge commit? Couldn't they just report coverage on the last commit of a PR? I.e. on 74f60a6?

mbillingr:
I assume that is the only commit ci/codecov gets to see. If you look at the travis details, it also says Commit 9613d8f.

cbrnr:
I see. So I guess we're happy with the new coverage? It would be great if the comment could be shortened (#158), but apparently this is not possible. So +1 for _not_ merging and closing this PR :-).

PR #158, opened 2016-03-15
Codecov: Removed suggestions

mbillingr:
> @mbillingr can you try disabling suggestions in codecov.yml line 22?

@cbrnr OK. Why?

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -19,5 +19,5 @@ coverage:
             if_ci_failed: error
 
 comment:
-    layout: "header, diff, changes"
+    layout: ""
     behavior: default
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -20,4 +20,4 @@ coverage:
 
 comment:
     layout: "diff, changes"
-    behavior: default
+    behavior: spammy
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -19,5 +19,5 @@ coverage:
             if_ci_failed: error
 
 comment:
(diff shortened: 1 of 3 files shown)


cbrnr:
I was wondering if the comment could be made more compact somehow. Obviously, removing suggestions didn't change anything. Do you know if the format of the comment can be changed somehow? In particular, the last 2 lines below the table could be removed.

mbillingr:
Maybe it *used* to look like that?

I think the configuration applies only to the verbatim part of the post.

cbrnr:
Maybe. Can you try to change some parameters in this line to see if it has any effect?

cbrnr:
So... why is the coverage check not shown?

mbillingr:
because argh!?

cbrnr:
Either it's a problem with GitHub, or Codecov is not really better than Coveralls...

cbrnr:
Also, if this ever works, maybe this is a nice plugin: https://github.com/codecov/browser-extension#codecov-extension

mbillingr:
The browser extension works nicely. I don't know how it's done, but if I browse to a source file in the master branch I get nice green/yellow/red line backgrounds

mbillingr:
Apparently, codecov did not like the empty layout format string.
And, apparently it does not update the comment if the coverage does not change...

cbrnr:
Since you've created a separate repo for testing such issues, I'm closing this one.

Issue #160, opened 2016-03-17
Stability of MVARICA


cbrnr:
On some of my data sets, I get different ICA decompositions depending on the Python distribution (Anaconda vs. Homebrew). Using a brewed Python 3.5.1, I even get different decompositions when I run the test script from the command line vs. from an interactive session. Of course I've set the random state to rule out differences due to different random numbers.

The decompositions differ not only in their order, but also in their morphology.

I wasn't able to come up with a simple test script to demonstrate this behavior. I do have a script which uses one of my data sets where this problem occurs. I could upload it if needed.

Note that this problem occurs only when using Infomax (we're using the MNE implementation) - FastICA from sklearn always yields identical results. Furthermore, this problem only occurs in combination with MVARICA (and likely CSPVARICA) - plain Infomax ICA is not affected. Therefore, I suspect it has something to do with applying ICA to the residuals of the VAR model (which might sometimes get very small). @mbillingr, what if some properties of the residuals contradict some assumptions of ICA? E.g., ICA needs Gaussian mixtures to decompose, and VAR residuals should be white and Gaussian?

mbillingr:
Even if assumptions are contradicted in some way I would expect identical results. 

The problem, as you discribe it could be caused at to places: ICA or VAR fitting. Let's try toisolate it. Could you please store the residuals on one distribution and run ICA on them in another distribution?
If there is a difference in the decomposition we know for sure it's the ICA.

cbrnr:
Today, I re-installed my brewed Python and the problem is gone. Which is good, because now the results are identical, no matter which Python I use or if I run the script directly or interactively. I still get this warning message when running from the brewed Python:

```
/usr/local/lib/python3.5/site-packages/scipy/linalg/basic.py:884: RuntimeWarning: internal gelsd driver lwork query error, required iwork dimension not returned. This is likely the result of LAPACK bug 0038, fixed in LAPACK 3.2.2 (released July 21, 2010). Falling back to 'gelss' driver.
  warnings.warn(mesg, RuntimeWarning)
```

However, this doesn't seem to have an effect on the results. Weird! I spent a lot of time hunting down the differences in components, and now everything works. I'll double check again to make sure I'm not crazy, but I guess we can close this issue...

cbrnr:
OK, let's not close this issue yet. Although I can't reproduce the differences in my test script, I still see them in my analysis code.

[...]

PR #161, opened 2016-03-22
Change ICA default to extended Infomax

### LINKED ISSUE #53: Use extended Infomax by default
As mentioned in #49, SCoT should use extended Infomax by default (just like EEGLAB does).

cbrnr:
Fixes #53. I think standard Infomax has problems with estimating sources as discussed in #160. Although I'm not sure that the toy example shown there is related to differences in components obtained with Anaconda vs. Homebrew Python, extended Infomax is the default choice for anyone applying ICA to EEG data. Therefore, we should use it too.


mbillingr:
Looks goot to me.

PR #165, opened 2016-03-30
Add MNE backend (Infomax and CSP)

### LINKED ISSUE #163: We need a regularized covariance estimator
For some data sets, CSP fails in the eigenvalue decomposition step. This is likely because the covs are not positive definite and therefore we need to regularize the covariance matrices (see discussion at https://github.com/mne-tools/mne-python/issues/2437).

@mbillingr, how should we go about and implement Ledoit-Wolf shrinkage? I would like to use the `sklearn` implementation, but `sklearn` is only an optional dependency.

cbrnr:
Fixes #163

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,21 +14,21 @@ env:
     - DISTRIB="conda" PYTHON_VERSION="2.7" USE_MKL="false"
       NUMPY_VERSION="1.8.2" SCIPY_VERSION="0.13.3"
       SKLEARN_VERSION="0.15.0" MATPLOTLIB_VERSION="1.4.0"
-      MNE="0.11.0" INSTALL_SCOT="true" RUN_EXAMPLES="false"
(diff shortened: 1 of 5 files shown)


cbrnr:
@mbillingr what do you think?

mbillingr:
The implementation looks good, but I am concerned about testing.

We may accidentally introduce code that depends on MNE outside of the backend, but our tests wouldn't catch that problem because MNE is available in all testing environments. Maybe it would make sense to have an environment that checks only the minimum dependencies...

However, the same issue exists for scikit-learn. So I guess we should address this somewhere else.

:+1:  for merge

cbrnr:
+1 for a minimal testing env without the optional dependencies.

So only numpy and scipy, and not matplotlib, scikit-learn, and mne. BTW, it's not just scikit-learn, many of our tests already depend on matplotlib, which we do not list as a required dependency.

PR #168, opened 2016-04-07
Use regularized covariance in CSP by default

cbrnr:
I think it doesn't hurt to use regularization by default. At least with my data, I need to use it because otherwise `eigh` fails.


cbrnr:
@mbillingr are you OK with this change?

mbillingr:
:+1:

PR #169, opened 2016-04-07
Implement basic Workspace __repr__

### LINKED ISSUE #124: Add __print__ (or __repr__) to Workspace class

cbrnr:
Fixes #124 

@mbillingr, is there an easier way to get the name of the current backend (other than what I did)?

### DIFF
--- a/scot/ooapi.py
+++ b/scot/ooapi.py
@@ -76,7 +76,13 @@ class Workspace(object):
         self.locations_ = locations
         self.reducedim_ = reducedim
         self.nfft_ = nfft
-        self.backend_ = backend if backend is not None else global_backend
+
+        if backend is None:
+            self.backend_ = global_backend
+        elif isinstance(backend, str):
+            self.backend_ = global_backend.get_backend(backend)
+        else:
+            self.backend_ = backend
 
         self.trial_mask_ = []
 
--- a/scot/ooapi.py
+++ b/scot/ooapi.py
@@ -100,6 +100,12 @@ class Workspace(object):
         except TypeError:
             self.var_ = var
 
+    def __repr__(self):
+        s = ("Workspace(var={}, locations={}, reducedim={}, nfft={}, fs={}, "
+             "backend={})".format(self.var_, self.locations_, self.reducedim_,
+                                  self.nfft_, self.fs_, self.backend_))
+        return s
+
     def __str__(self):
         if self.data_ is not None:
             datastr = '%d trials, %d channels, %d samples' % self.data_.shape
--- a/scot/backendmanager.py
+++ b/scot/backendmanager.py
@@ -9,6 +9,10 @@ class BackendManager:
     def __init__(self):
         self.backends = {}
         self.current = None
+        self.current_name = None
+
+    def __repr__(self):
+        return self.current_name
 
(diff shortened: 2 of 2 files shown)


cbrnr:
The output currently looks like this:

    Workspace(var=<scot.var.VAR object at 0x10cc865c0>, locations=None, reducedim=None, nfft=512, fs=2, backend=builtin)

mbillingr:
>  is there an easier way to get the name of the current backend (other than what I did)?

No, looks fine. There are other ways but I think they are more complicated.

> (I think a backend object has to be supplied). How do we handle this?

I think `self.current_name` is not an accurate representation of the BackendManager. What about `str(self.current)`?

cbrnr:
> I think `self.current_name` is not an accurate representation of the BackendManager. What about `str(self.current)`?

Makes things less easy to read IMO. It would be nice if we could set the backend with a string argument.

mbillingr:
> Makes things less easy to read IMO. It would be nice if we could set the backend with a string argument.

Why not? We could do this in the workspace:
```Python
if backend is None:
    self.backend_ = global_backend
elif isinstance(backend, str):
    self.backend_ = global_backend.get_backend(backend)
else:
    self.backend_ = backend
```

cbrnr:
So you think it is then OK to represent a backend with its name?

cbrnr:
This doesn't work as expected - if a backend is supplied, `global_backend.get_backend(backend)` doesn't return a `BackendManager` object, so the representation is different and doesn't show the `current_name`.

mbillingr:
> So you think it is then OK to represent a backend with its name?

It's not nice, but I don't have a better idea.

scot/ooapi.py
```diff
@@ -100,6 +106,12 @@ def __init__(self, var, locations=None, reducedim=None, nfft=512, fs=2, backend=
         except TypeError:
             self.var_ = var
 
+    def __repr__(self):
+        s = ("Workspace(var={}, locations={}, reducedim={}, nfft={}, fs={}, "
+             "backend={})".format(self.var_, self.locations_, self.reducedim_,
+                                  self.nfft_, self.fs_, self.backend_))
```
mbillingr: if you do `repr(self.backend_)` we will get the correct string representation.

cbrnr: No, we don't, because `global_backend.get_backend(backend)` returns a `dict` and not a `BackendManager` object.

mbillingr: This is a problem because the name is separated from the backend. We could make a cusom class (derived from `dict`) with a `name` property.

cbrnr: Or just add a new dict entry `"name": "sklearn"`?

cbrnr: I don't like the fact that a `dict` is returned if we specify a backend with a string, whereas otherwise a `BackendManager` is required. Do you have a suggestion how to streamline this?

mbillingr: We can store the backend argument and use this for representation.

cbrnr:
Closing, if we ever need this, it is probably easier to start from scratch.

PR #172, opened 2016-06-10
Nicer plot_topo

cbrnr:
Options for plotting the locations and the head (default plot both) and a nicer look by default (smaller electrode dots with transparency, thinner lines).

### DIFF
--- a/scot/eegtopo/topoplot.py
+++ b/scot/eegtopo/topoplot.py
@@ -184,7 +184,8 @@ class Topoplot(object):
         return self.channel_fence
 
 
-def topoplot(values, locations, axes=None, offset=(0, 0), **kwargs):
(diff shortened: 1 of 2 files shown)


cbrnr:
@mbillingr and once again our tests fail. Did we break anything or is it Travis, Anaconda or a combination?

cbrnr:
And yes, in case you were wondering, all tests pass on my local machine.

cbrnr:
See #173 for Travis issues

mbillingr:
Hmm.. is there any sensible way of testing plotting functions other than trying to run the code and see if there are errors?

cbrnr:
I can't think of something else, because we don't want to test for specific elements in the plot, do we?

cbrnr:
@mbillingr OK for merge?

PR #173, opened 2016-06-13
Fix Travis setup

cbrnr:
Trying to fix Travis in this PR...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,24 +11,27 @@ virtualenv:
 env:
   matrix:
     # oldest supported package versions on Python 2
-    - DISTRIB="conda" PYTHON_VERSION="2.7" USE_MKL="false"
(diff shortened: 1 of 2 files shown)


cbrnr:
Looks like the old package versions aren't compatible with Python 3.5 anymore. We could use Python 3.4, or we could just test the old versions with Python 2.7.

cbrnr:
I'd go for the current setup:

- Python 2.7 and oldest supported packages
- Python 2.7 with current versions
- Python 3.5 with current versions

Current versions as available in Anaconda. @mbillingr what do you think?

cbrnr:
I've merged this because otherwise our tests are broken which affects all other PRs.

mbillingr:
Ok, this is fine with me :)

PR #174, opened 2016-06-14
Fix wrong axis argument when removing activations

cbrnr:
I think the axis argument is wrong - we want to delete components, which I think correspond to the rows. @mbillingr, can you confirm? And if this is really a bug, why didn't we notice this before (e.g. in our tests)?


cbrnr:
Ah, forget it! I was testing this with a 2D array, whereas it should always be a 3D array! Therefore, in my case it was axis=0, and in the 3D case it was axis=1.

PR #175, opened 2016-06-14
Add MNE >= 0.11.0 as optional dependency

cbrnr:
@mbillingr please merge if you are OK with these changes.


mbillingr:
Do we use MNE for anything in our code?

cbrnr:
We use CSP and Infomax from MNE in our MNE backend.

PR #176, opened 2016-06-15
Nicer plot_connectivity_spectrum

cbrnr:
Compare the two figures (top: current, bottom: new):
figure_current
figure_new
The topos have also changed, but they're not part of this PR. I'll integrate them in #172.


mbillingr:
This looks good. What about the y-axis?

cbrnr:
I thought the y-axis scaling is not really important, especially since it doesn't seem to be normalized in any way (at least I think that's the case for the ffDTF). So as long as all plots are scaled the same way the exact values are not that important. What do you think?

PR #178, opened 2016-08-18
Clean up PCA code

cbrnr:
Clean up docstrings and generally make code a bit nicer.


cbrnr:
@mbillingr these are mostly minor stylistic changes, but I think the argument `full_matrices=False` is necessary for `svd` to work.

PR #182, opened 2016-09-14
Implement data loader and downloader

cbrnr:
Not all examples are adapted yet. Supersedes #128.

### DIFF
--- a/scot/datasets.py
+++ b/scot/datasets.py
@@ -6,6 +6,7 @@ from os import makedirs
 from os.path import expanduser, isfile, isdir, join
 from requests import get
 import numpy as np
+import hashlib
 
 from .matfiles import loadmat
 from .eegtopo.eegpos3d import positions
@@ -13,7 +14,8 @@ from . import config
 
 
 datadir = expanduser(config.get("scot", "data"))
-datasets = {"mi": {"files": ["motorimagery.mat"], "url": "https://github.com/scot-dev/scot-data/raw/master/scotdata/"}}
+datasets = {"mi": {"files": ["motorimagery.mat"], "md5": ["239a20a672f9f312e9d762daf3adf214"],
+                   "url": "https://github.com/scot-dev/scot-data/raw/master/scotdata/"}}
 
 
 def fetch(dataset="mi", datadir=datadir):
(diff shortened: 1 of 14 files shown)


cbrnr:
@mbillingr any chance you can take a look and let me know what you think?

cbrnr:
Failing because of an Anaconda bug https://github.com/ContinuumIO/anaconda-issues/issues/1068, I hope they'll fix that soon...

cbrnr:
Nice, working again. @mbillingr what do you think? IMO we could replace the `scot-data` repo with this implementation now...

scot/datasets.py
```diff
@@ -0,0 +1,85 @@
...
+
+from .matfiles import loadmat
+from .eegtopo.eegpos3d import positions
+from . import config
+
+
+datadir = expanduser(config.get("scot", "data"))
+valid_datasets = ["mi"]
+files = {"mi": ["motorimagery.mat"]}
+url = {"mi": "https://github.com/scot-dev/scot-data/raw/master/scotdata/"}
```
mbillingr: Instead of multiple dicts I would prefer something like `datasets = {"mi": {"files": ..., "url": ...}}`.

Both approaches have their benefits. I'd just prefer not having to edit multiple placed when adding another dataset in the future.

scot/datasets.py
```diff
@@ -0,0 +1,85 @@
+# Released under The MIT License (MIT)
+# http://opensource.org/licenses/MIT
+# Copyright (c) 2016 SCoT Development Team
+
+from os import makedirs
+from os.path import expanduser, isfile, isdir, join
+from requests import get
```
mbillingr: Is requests a common enough library that depending on it is not too troublesome for potential users?

scot/datasets.py
```diff
@@ -0,0 +1,85 @@
...
+              "locations" ... Channel locations
+    """
+    if dataset not in valid_datasets:
+        raise ValueError("Example data '{}' not available.".format(dataset))
+    if not isdir(datadir):
+        makedirs(datadir)
+
+    data = []
+
+    for file in files[dataset]:
```
mbillingr: In python 2 `file` is a built-in function. It's probably bad karma to overwrite it :)

scot/datasets.py
```diff
@@ -0,0 +1,85 @@
...
+            The data set is stored in a list, where each list element
+            corresponds to data from one subject. Each list element is a
+            dictionary with the following keys:
+              "eeg" ... EEG signals
+              "triggers" ... Trigger latencies
+              "labels" ... Class labels
+              "fs" ... Sample rate
+              "locations" ... Channel locations
+    """
+    if dataset not in valid_datasets:
```
cbrnr: Of course! Much better!

scot/datasets.py
cbrnr: Didn't know that since I don't care about Python 2 :-) - but good to be compatible, I'll change it to the proven and tested name `f`.

cbrnr: `f` is already taken, so I'll use `filename` instead.

cbrnr:
Except for the `requests` issue this is it. If you want to use something more standard, it would be great if you could directly provide the alternative code.

cbrnr:
@mbillingr I added an MD5 check - OK?

mbillingr:
MD5 sounds great!

PR #203, opened 2023-04-20
Update README and remove Python 2 support

cbrnr:
Let's see if I can take this package to the present. This is a small start.


cbrnr:
OK OK. This won't be that easy. I cannot even install the package in editable mode locally. Tests need to be converted to `pytest`. Travis needs to be replaced with GitHub Actions.

mbillingr:
I exclusively use [poetry](https://python-poetry.org/) now to manage dependencies and local development of Python packages. It uses `pyproject.toml` instead of `setup.py`/`setup.cfg`, which I find much more intuitive to maintain.

What do you think of switching?

mbillingr:
> [pip] also supports pyproject.toml nowadays

nice! :)

PR #204, opened 2023-05-04
Modernize tests (switch to pytest)

mbillingr:
The good news: pytest works out of the box!

Using Python 3.10.something I installed recent versions of our dependencies, ran pytest, and it worked. Well, mostly... I had to fix a few uses of floats and adjust to changes in the APIs of a few dependencies.

TODO
- [x] Fix ~2000 warnings (many deprecations), maybe in separate PR?
- [ ] Use pytest's flat test functions?
  - @cbrnr  Can you remind me what's wrong with test classes? Tests seem nicely structured that way.

### DIFF
--- a/scot/backend_mne.py
+++ b/scot/backend_mne.py
@@ -11,6 +11,11 @@ from . import datatools
 from . import backend
 from . import backend_builtin as builtin
 
+try:
+    import mne
+except ImportError:
+    mne = None
+
 
 def generate():
     from mne.preprocessing.infomax_ import infomax
@@ -36,4 +41,5 @@ def generate():
     return backend
 
 
-backend.register('mne', generate)
+if mne is not None:
+    backend.register('mne', generate)
--- a/scot/backend_sklearn.py
+++ b/scot/backend_sklearn.py
@@ -5,7 +5,7 @@
 """Use scikit-learn routines as backend."""
 
 from __future__ import absolute_import
-import scipy as sp
+import numpy as np
 
 from .datatools import atleast_3d, cat_trials, dot_special
 from . import backend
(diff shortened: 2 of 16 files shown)


mbillingr:
Note that I disabled the travis integration for now, so that we are able to merge pull requests again.

cbrnr:
> Can you remind me what's wrong with test classes? Tests seem nicely structured that way.

There's nothing wrong with test classes, but I'd rather use functions for the following two reasons:

1. Certain features like `@pytest.mark.parametrize` only work with functions (I think).
2. Most people seem to be using test functions nowadays, so to minimize future efforts I think it would be nice if we did that here as well.

That said, it's more important to get our tests to run (on GitHub Actions) again, so we can always postpone the switch to test functions to a follow up PR.

> Fix ~2000 warnings (many deprecations), maybe in separate PR?

If it's not too much work I'd do it right here.

mbillingr:
> That said, it's more important to get our tests to run (on GitHub Actions) again, so we can always postpone the switch to test functions to a follow up PR.

Agreed, let's save that for later.

> If it's not too much work I'd do it right here.

It was not that much work, I think warnings were repeated a lot.

mbillingr:
image
:partying_face:

cbrnr:
A bunch of warnings though...

cbrnr:
Also, there's a warning about too many open figures. Do we not close figures in our tests?

mbillingr:
> A bunch of warnings though...

I did not see those locally. Hopefully my fix did not break anything.


> Also, there's a warning about too many open figures. Do we not close figures in our tests?

We do:
```python
    def tearDown(self):
        plt.close('all')
```

I remember this warning, though. Maybe we never really solved this.

cbrnr:
> We make extensive use of `self.assertSomething`, which need to be rewritten.

That should be just a simple `assert` statement or some `numpy` variant.

> Also, we sometimes precompute slow operations in `setUp` for reuse in multiple tests. I'm not sure how easy it is to translate this pattern into pytest fixtures.

If it's just computed once, i.e. not for every test, I think you can include relevant code at the top of the test module (in global namespace), since this should be executed first.

> I wonder if the effort would be worth the gain...

For me, yes, definitely. In all projects I'm involved we're using test functions.

cbrnr:
Plot warnings are gone!

cbrnr:
@mbillingr ready for merge?

cbrnr:
Thanks @mbillingr!

PR #205, opened 2023-05-04
add a github action

mbillingr:
I'm quickly adding an action that runs pytest, so we get some PR safety back.

We can later replace this with better actions.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -1,0 +1,40 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: Python package
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


cbrnr:
👍 – can you add Python 3.11 as well?

mbillingr:
Failing tests will be fixed in #204.

cbrnr:
> Sure. What's the lowest version we want to support?

I'd say 3.8, or if you want even 3.9. Many packages are dropping 3.8 support now.

cbrnr:
Also, I'd rather not skip versions in between. We can only claim that we support those versions if we test explicitly.

mbillingr:
Right. I'll drop 3.8 then, so we don't have to take action when our dependencies no longer support it :)

mbillingr:
Can we merge this? I'd like the action to run on #204.

cbrnr:
Thanks @mbillingr!

PR #206, opened 2023-05-04
Use main instead of master

cbrnr:
This shouldn't break anything, but let's see.


mbillingr:
I'm positively surprised that active PRs were adjusted to merge with main.

PR #209, opened 2024-02-20
make documentation buildable in 2024 :)

mbillingr:
I managed to build the documentation again! :partying_face: 

Various incompatibilities had to be fixed, but i am amazed how well it still works after 10 years.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,3 +37,21 @@ To run the examples on Linux, invoke the following commands inside the SCoT dire
 etc.
 
 Note that the example data from https://github.com/SCoT-dev/scot-data needs to be available. The `scot-data` package must be on Python's search path.
+
+
+### Building the docs
+
+In February 2024 we managed to build the docs with the following package versions:
+
+```
+[tool.poetry.dependencies]
+python = "^3.11"
+sphinx = "^7.2.6"
+matplotlib = "^3.8.3"
+scipy = "^1.12.0"
+scikit-learn = "^1.4.1.post1"
+```
+
+Note that these are the most recent versions at the moment, so it is likely that future versions will just work.
+When using a newer version of sphinx, it may be necessary to update the subrepository in doc/sphinxext/numpydoc.
+
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,8 @@
+# Updating the docs
+
+1. run `generate_apidoc.sh`
+2. run `make.py`
+3. fix any errors and repeat steps 1-2 as often as necessary
+4. (ignore that building the latex docs fails)
+5. copy the content of build/html into the scot-doc repository, commit, push
+
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
 #
 # SCoT documentation build configuration file, created by
 # sphinx-quickstart on Thu Jan 23 12:52:18 2014.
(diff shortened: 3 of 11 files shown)


cbrnr:
Oh wow, thanks @mbillingr! Have you already updated the docs so that the jquery-related security issue is resolved? Are there any packages pinned, or does the build always use the latest available versions? And finally, why are the tests failing :laughing:?

mbillingr:
> Have you already updated the docs so that the jquery-related security issue is resolved?

Yes, I've updated the doc pages but I don't know if this issue was resolved. Dependabot will tell us next week, I guess :)


> Are there any packages pinned, or does the build always use the latest available versions?

Nothing pinned. I just installed the newest version of each missing package. If there were any issues I tried to fix them in our sources. (Actually, I think I broke the LaTeX build but I don't see why we would need it anyway.) The closest thing we have to a pinned dependency is the numpydoc sphinx extension subrepository. I've upgraded it to the newest version without any problems, though.


>  And finally, why are the tests failing 😆?

No idea. They are failing for something I did not even touch in this PR... how annoying!

cbrnr:
> Nothing pinned. I just installed the newest version of each missing package. If there were any issues I tried to fix them in our sources. (Actually, I think I broke the LaTeX build but I don't see why we would need it anyway.) The closest thing we have to a pinned dependency is the numpydoc sphinx extension subrepository. I've upgraded it to the newest version without any problems, though.

Can you maybe add a sentence on how to update packages to the instructions? Ideally, the instructions should be self-contained so that we can re-build the docs again in 2034!

> No idea. They are failing for something I did not even touch in this PR... how annoying!

:smile: That's not the first time I've seen this happen! How do you want to proceed? Do you have time to take a look? Or should we go ahead and merge with the failing tests :fearful:?

mbillingr:
> Can you maybe add a sentence on how to update packages to the instructions?

I started with a fresh venv and did a `pip install` for whatever was raising an import error. Do you think that information would be useful in the documentation?


> 😄 That's not the first time I've seen this happen! How do you want to proceed? Do you have time to take a look? Or should we go ahead and merge with the failing tests 😨?

The bad news is that the tests are broken, but the good news is that the tests are broken. :)
I quickly checked. We are comparing floats for equality, which fails because of rounding errors. I can fix the failing test but I don't have the time to look for all comparisons.

cbrnr:
> I started with a fresh venv and did a `pip install` for whatever was raising an import error. Do you think that information would be useful in the documentation?

Could we not put the required dependencies in a dedicated optional dependencies group?

> The bad news is that the tests are broken, but the good news is that the tests are broken. :)
I quickly checked. We are comparing floats for equality, which fails because of rounding errors. I can fix the failing test but I don't have the time to look for all comparisons.

OK, when did this start happening? Let's merge this doc PR once we've clarified how we want to write the instructions, and then maybe just open a new issue for the failing tests?

mbillingr:
> OK, when did this start happening? 

We probably always had them and they occasionally start failing when we update numpy, scipy, or sklearn.


> Could we not put the required dependencies in a dedicated optional dependencies group?

We have neither a requirements.txt nor a pyproject.toml, so I'm not sure where to create a dependency group... actually, I could not find a single dependency specififacion in any of our setup files.
Modernizing the build system is a can of worms I don't want to open right now, but I could add the library versions I used to the doc/README.md. What do you think?

mbillingr:
>  preferably directly in README.md

Done.

cbrnr:
Perfect, thank you so much!

PR #210, opened 2024-02-20
fix floatingpoint equality checks in tests

mbillingr:
Why are we still checking floats for equality? :smile:


cbrnr:
Thanks @mbillingr!

PR #211, opened 2025-04-02
Changes to get PDC method in connectivity to work

ddsou:
Basically to run something simple like
data = np.stack([x0, x1, x2], axis=1).reshape(1, -1, 3)
model = VAR(3) 
model = model.fit(data)
conn = Connectivity(model.coef, nfft=10)
conn_matrix = conn.PDC()

I had to change an sp depreciation and adjust _construct_var_eqns to give me a proper 3x3 matrix at the end

Very possible I've used it wrong but thought I'd post this anyways :)

### DIFF
--- a/scot/var.py
+++ b/scot/var.py
@@ -72,7 +72,7 @@ class VAR(VARBase):
             # regularized least squares (ridge regression)
             x, y = self._construct_eqns_rls(data)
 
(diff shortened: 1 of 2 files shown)


cbrnr:
Thank you, I'll take a look soon!

